PR #9, opened 2026-06-08
Implement StatBlockGenerator v2 draft API (command-board contract)

Drakosfire:
### Motivation

- Provide a narrow v2 producer contract so DungeonBuddy can request reviewed live-combat statblock drafts with structured data, markdown, deterministic combat defaults, warnings, provenance, and a clear lifecycle state. 
- Keep the existing app-facing generation API intact while adding a safe, typed adapter layer that reuses the existing generator core rather than changing generator internals. 

### Description

- Added v2 command-board Pydantic models in `statblockgenerator/models/command_board_contract_models.py` including `StatBlockDraftRequest`, `StatBlockDraft`, `CombatDefaults`, `DraftProvenance`, `ReviewWarning`, and a stable `StatBlockDraftResponse`/`ContractError` envelope. 
- Implemented draft adapter helpers in `statblockgenerator/services/statblock_draft_adapter.py` that convert the v2 request to `CreatureGenerationRequest`, render conservative markdown, derive deterministic `combat_defaults`, extract `save_dcs`, build lightweight `warnings`, and assemble the v2 `StatBlockDraft` with provenance and `persist` defaulting to `false`. 
- Exposed a canonical v2 health route and a generate-draft route by adding `GET /api/statblockgenerator/v2/health` and `POST /api/statblockgenerator/v2/generate-draft` to `routers/statblockgenerator_router.py`, returning `501` for accepted-but-not-implemented revise/source modes and mapping generator failures to the stable error envelope. 
- Exported the new contract models from `statblockgenerator/models/__init__.py` and added unit tests for contract model validation, adapter behavior, and route behavior in `tests/statblockgenerator/test_command_board_contract_models.py`, `tests/statblockgenerator/test_statblock_draft_adapter.py`, and `tests/statblockgenerator/test_statblockgenerator_v2_routes.py` (route tests mock `StatBlockGenerator.generate_creature()` so they do not call OpenAI). 

### Testing

- Performed static checks by compiling the new/changed modules with `python -m py_compile ...` which succeeded for the added files. 
- Attempted to run the unit test suite with `python -m pytest tests/statblockgenerator/test_command_board_contract_models.py -v`, `python -m pytest tests/statblockgenerator/test_statblock_draft_adapter.py -v`, and `python -m pytest tests/statblockgenerator/test_statblockgenerator_v2_routes.py -v`, but test execution was blocked by an environment dependency fetch error for the external `GenerationEngine` dependency (network/Git fetch returned a CONNECT tunnel 403), so the pytest runs could not complete in CI here. 
- Tests are written to avoid live OpenAI calls (route tests mock `generate_creature()`); to validate locally, run the pytest commands above in an environment where `generationengine` can be fetched or vendored and dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2639278520832090325abb54c5b930)

### DIFF
--- a/routers/statblockgenerator_router.py
+++ b/routers/statblockgenerator_router.py
@@ -34,11 +34,22 @@ import os
 import fal_client
 from openai import OpenAI
 from cloudflare.handle_images import upload_image_to_cloudflare
+from fastapi.responses import JSONResponse
+
+from statblockgenerator.models.command_board_contract_models import (
+    ContractError,
+    StatBlockDraftRequest,
+    StatBlockDraftResponse,
+)
+from statblockgenerator.services.statblock_draft_adapter import (
+    build_draft,
+    build_generation_request,
+)
 
 # NOTE: Image generation now handled by /api/images/generate (image_management_router.py)
 
 # Initialize OpenAI client (still used for text generation)
-openai_client = OpenAI()
+openai_client = OpenAI(api_key=os.getenv("OPENAI_API_KEY")) if os.getenv("OPENAI_API_KEY") else None
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


Drakosfire (commented):
Reviewed. No blocking changes from my side.

This lands the intended narrow seam: v2 producer models, a boring adapter around the existing generator core, canonical v2 health, mocked route coverage, deterministic combat defaults, markdown rendering, warnings, and provenance. It keeps the existing `/api/statblockgenerator/generate-statblock` endpoint stable, which was the main safety boundary.

Strengths:
- Good separation between contract models, adapter helpers, and router logic.
- `combat_defaults` are derived from `StatBlockDetails`, not separately invented by the LLM.
- The response envelope is stable enough for DungeonBuddy to start building a mock/real client against it.
- Returning 501 for accepted-but-unimplemented source/revision modes is a good explicit boundary for this slice.
- Route tests mock `generate_creature()`, keeping CI free of live OpenAI calls.

Non-blocking follow-ups I would consider before/after merge:
- Add a small model validator on `StatBlockDraftResponse` so `success=True` requires `draft` and `success=False` requires `error`. The route currently behaves correctly, but the contract model does not enforce it.
- Only emit `missing_markdown` when `output_options.include_markdown` is true. Right now `include_markdown=false` would intentionally produce empty markdown but still create an error warning.
- If `include_review_warnings=false`, consider whether `review_status="warnings"` with an empty warnings list is desirable; it may be confusing to consumers.
- Add a route test for the 501 path so the deferred `generate_from_source_statblock` / `revise_existing` behavior is pinned.
- Longer term: make fixture paths relative to the test file or assert that fixture discovery is non-empty, so tests do not silently skip if run from a different working directory.

Given the scope and the dependency-fetch issue noted in the PR description, I’d treat this as mergeable after local pytest validation in an environment that can install the external dependency.

PR #10, opened 2026-06-08
Add StatBlockGenerator v2 `render-draft` endpoint

Drakosfire:
### Motivation
- Provide a simple v2 producer path that accepts an existing `StatBlockDetails` payload and returns the command-board draft envelope without invoking the generator core. 
- Surface `render-draft` capability in the v2 health discovery so consumers know the adapter can wrap authored or previously-generated statblocks. 
- Keep the v2 contract orthogonal to the existing app-facing `generate-statblock` endpoint while enabling DungeonBuddy to render drafts for accept-to-combat workflows.

### Description
- Add `StatBlockDraftRenderRequest` Pydantic model to `statblockgenerator/models/command_board_contract_models.py` to represent requests that supply an existing `StatBlockDetails` payload. 
- Export the new `StatBlockDraftRenderRequest` from `statblockgenerator/models/__init__.py`. 
- Add `POST /api/statblockgenerator/v2/render-draft` in `routers/statblockgenerator_router.py` which wraps the provided statblock with `build_draft(...)`, sets provenance `generation_info` to `{"source": "render-draft", "generated": False}`, and returns the stable `StatBlockDraftResponse` envelope without calling `generate_creature`. 
- Update the v2 health route to advertise `"supports": ["generate-draft", "render-draft"]` and add route-level tests to assert behavior and that generation is not invoked.

### Testing
- Ran static compile checks with `python -m py_compile` on the updated modules, which succeeded. 
- Added/updated route tests in `tests/statblockgenerator/test_statblockgenerator_v2_routes.py` to include `test_render_draft_wraps_existing_statblock_without_generation`, and the test logic validates markdown, combat defaults, provenance, and that `generate_creature` is not awaited. 
- Running `pytest` in the current environment is blocked: test execution failed with `ModuleNotFoundError: No module named 'dotenv'` and environment tooling attempts (`uv` / alternative Python runtimes) were blocked by platform wheel/availability issues (torch wheel mismatch for the active interpreter and a download 403 when trying to provision alternate runtimes).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a263f46bc2c8320bda52ef78d27df6c)


Drakosfire (commented):
Reviewed and critiqued.

The endpoint is useful and directionally right: `render-draft` gives DungeonBuddy a cheap way to wrap authored / previously generated statblocks into the same command-board draft envelope without invoking generation. That is a good complement to `generate-draft`, especially for Statblock View and accept-to-combat flows.

The main issue I would fix before leaning on this contract: provenance currently has mixed semantics. `render-draft` builds an adapter request with `mode="generate_from_source_statblock"`, while `generate-draft` explicitly returns 501 for that same mode. The resulting draft provenance therefore says `mode: generate_from_source_statblock`, even though no generation happened. Also, `DraftProvenance.generator` still defaults to `StatBlockGenerator.generate_creature`, while this route explicitly does not call `generate_creature`. The `generation_info` payload does say `{source: "render-draft", generated: false}`, but that is compensating for two misleading top-level provenance fields.

Recommended fix: add an explicit render/wrap mode to the contract, for example `render_existing` or `render_draft`, and allow `build_draft` / `DraftProvenance` to record the actual adapter/source, e.g. `generator: "statblock_draft_adapter.render_existing"` or `producer: "render-draft"`. This makes downstream audit much cleaner for DungeonBuddy.

Other non-blocking notes:
- `include_combat_defaults` still is not honored because `combat_defaults` is always computed and required in the response. That may be fine for now, but the option should either be removed/deferred or documented as always-on for v0.
- A test should assert that the render response provenance does not claim the generator core was invoked. Right now it only checks `generation_info` and `mock_generate.assert_not_awaited()`.
- The route is intentionally narrow, which is good. I would not add revision/generation logic here.

Net: useful PR, but I would strongly prefer the provenance/mode cleanup before treating `render-draft` as a durable consumer contract.

Drakosfire (commented):
```markdown
Reviewed PR #10 re-review.

Result: mergeable from design/contract review, with local test validation still needed.

Summary:
- The revised PR fixes the main provenance concern from the first review.
- `render-draft` now uses an explicit `render_existing` mode instead of reusing `generate_from_source_statblock`.
- `build_draft(...)` now accepts a `generator` override, and `render-draft` records `generator: "statblock_draft_adapter.render_existing"` instead of implying `StatBlockGenerator.generate_creature` produced the draft.
- The route test now asserts both that `generate_creature()` is not awaited and that top-level provenance does not claim the generator core was used.
- `include_combat_defaults` is now documented as reserved/future behavior while combat defaults remain always emitted in v2, which is acceptable for this contract stage.
- Fixture discovery was made robust with a test-file-relative path and a non-empty fixture assertion.

Blocking changes:
- None from design/contract review.

Non-blocking follow-ups:
- GitHub currently reports the PR as `mergeable: false`. Confirm whether this is a real conflict, stale mergeability state, or branch update issue before merging.
- The PR body still says pytest was blocked by environment dependency issues. Before merge, run the focused tests locally in an environment with dependencies installed if possible:
  - `python -m pytest tests/statblockgenerator/test_command_board_contract_models.py -v`
  - `python -m pytest tests/statblockgenerator/test_statblock_draft_adapter.py -v`
  - `python -m pytest tests/statblockgenerator/test_statblockgenerator_v2_routes.py -v`
- Consider whether `render_existing` should eventually get its own `DraftMode` documentation in the audit/handoff docs, since Buddy clients will now see it in provenance.

Recommendation:
- The contract shape is now clean enough for DungeonBuddy to build against.
- Merge after resolving/confirming mergeability and validating the focused tests locally.
```

PR #11, opened 2026-06-08
Lock down StatBlockGenerator v2 internal endpoints

Drakosfire:
### Motivation

- The StatBlockGenerator v2 producer endpoints are intended for internal service consumers and must be protected from anonymous internet callers. 
- Add a narrow, shared-secret gate so internal product backends (e.g. DungeonBuddy) can call v2 routes server-to-server without exposing a secret to browsers. 

### Description

- Added a small reusable FastAPI dependency in `routers/internal_auth.py` that reads `DUNGEONBUDDY_INTERNAL_API_KEY` and validates the `X-DungeonBuddy-Internal-Key` header using `secrets.compare_digest`. 
- Applied the dependency to the v2 producer routes only: `GET /api/statblockgenerator/v2/health`, `POST /api/statblockgenerator/v2/generate-draft`, and `POST /api/statblockgenerator/v2/render-draft`, leaving legacy `POST /api/statblockgenerator/generate-statblock` unchanged. 
- Updated `routers/statblockgenerator_router.py` to import and attach `require_dungeonbuddy_internal_key` to those routes as route-level dependencies. 
- Added focused tests `tests/statblockgenerator/test_statblockgenerator_v2_auth.py` and updated `tests/statblockgenerator/test_statblockgenerator_v2_routes.py` to set the env var and send the internal header; tests cover missing/wrong/correct key behavior, missing server config, ensure generation/render logic is not invoked when auth fails, and assert the legacy route has no internal-key dependency. 

### Testing

- Performed static checks and compilation: `python -m py_compile routers/internal_auth.py routers/statblockgenerator_router.py tests/statblockgenerator/test_statblockgenerator_v2_auth.py tests/statblockgenerator/test_statblockgenerator_v2_routes.py` (succeeded). 
- Ran repository diff/checks for accidental changes (succeeded). 
- Attempted to run the focused tests with `python -m pytest tests/statblockgenerator/test_statblockgenerator_v2_auth.py tests/statblockgenerator/test_statblockgenerator_v2_routes.py -v` but the test run was blocked by the environment: `ModuleNotFoundError: No module named 'dotenv'` (missing `python-dotenv`). 
- Attempted a constrained `uv` run to use a different Python, but dependency sync failed while fetching a git-based dependency (`generationengine`) due to network restrictions (CONNECT tunnel 403), preventing full pytest execution in this environment. 

All new tests are included and validated by static checks; they exercise missing/wrong/correct key paths and ensure generation/render functions are not invoked when auth fails, but running them end-to-end requires installing test dependencies (`python-dotenv`) and allowing access to the `generationengine` dependency in CI or a developer environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a26ce83153c8320a4b79f0fd733d6d0)

### DIFF
--- a/routers/internal_auth.py
+++ b/routers/internal_auth.py
@@ -1,0 +1,28 @@
+"""Reusable internal service-to-service authentication dependencies."""
+
+import os
+import secrets
+from typing import Annotated
+
+from fastapi import Header, HTTPException
+
+INTERNAL_KEY_HEADER = "X-DungeonBuddy-Internal-Key"
+INTERNAL_KEY_ENV = "DUNGEONBUDDY_INTERNAL_API_KEY"
+
+
+async def require_dungeonbuddy_internal_key(
+    x_dungeonbuddy_internal_key: Annotated[
+        str | None,
+        Header(alias=INTERNAL_KEY_HEADER),
+    ] = None,
+) -> None:
+    """Require the DungeonBuddy internal API key for protected producer routes."""
+    expected_key = os.getenv(INTERNAL_KEY_ENV)
+    if not expected_key:
+        raise HTTPException(status_code=500, detail="Internal API key is not configured")
+
+    if x_dungeonbuddy_internal_key is None:
+        raise HTTPException(status_code=401, detail="Missing internal API key")
+
+    if not secrets.compare_digest(x_dungeonbuddy_internal_key, expected_key):
+        raise HTTPException(status_code=403, detail="Invalid internal API key")
--- a/routers/statblockgenerator_router.py
+++ b/routers/statblockgenerator_router.py
@@ -11,6 +11,7 @@ import time
 
 # Import authentication
 from .auth_router import get_current_user, get_current_user_optional
+from .internal_auth import require_dungeonbuddy_internal_key
 from auth_service import User
 
 # Import StatBlock components
@@ -82,7 +83,7 @@ async def health_check():
     }
 
 
-@router.get("/v2/health")
+@router.get("/v2/health", dependencies=[Depends(require_dungeonbuddy_internal_key)])
 async def v2_health_check():
     """Canonical health check for the command-board draft v2 contract."""
     return {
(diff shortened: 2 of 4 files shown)


Drakosfire (commented):
```markdown
Reviewed PR #11.

Result: mergeable from design/security review, with local test validation still needed.

Summary:
- This PR implements the intended narrow internal service-to-service gate for the StatBlockGenerator v2 producer endpoints.
- It adds `routers/internal_auth.py` with a reusable `require_dungeonbuddy_internal_key` dependency.
- It uses the agreed header/env contract: `X-DungeonBuddy-Internal-Key` checked against `DUNGEONBUDDY_INTERNAL_API_KEY`.
- It uses `secrets.compare_digest`, which is the right choice for shared-secret comparison.
- It protects only the v2 producer routes:
  - `GET /api/statblockgenerator/v2/health`
  - `POST /api/statblockgenerator/v2/generate-draft`
  - `POST /api/statblockgenerator/v2/render-draft`
- It leaves the legacy app-facing `/api/statblockgenerator/generate-statblock` route unchanged.
- The new auth tests cover missing key, wrong key, correct key, missing server config, generation not called when auth fails, render not called when auth fails, and legacy route dependency shape.
- Existing v2 route tests were updated to send the internal key, which keeps the contract tests aligned with the new security boundary.

Blocking changes:
- None from design/security review.

Non-blocking follow-ups:
- Local pytest still needs to be run in an environment with dependencies installed. The PR description notes `python-dotenv` / `generationengine` dependency issues blocked full execution in the current environment.
- Consider whether the 500 missing-env behavior should be monitored/logged at startup in a future deployment-hardening PR. Runtime fail-closed is acceptable for this slice.
- Consider documenting the header/env names in deployment docs or example env files when the Buddy proxy PR lands, so the two services stay aligned.
- Longer term, if more internal producer routes appear, consider grouping them under a protected sub-router instead of adding dependencies route-by-route. Route-by-route is fine for this small PR.

Recommendation:
- This is the right security hardening slice.
- Merge after focused local test validation, or merge with the known environment-test caveat if this repo’s current dependency setup makes local execution impractical.
```

PR #12, opened 2026-07-18
feat(statblocks_v1): PR12 bounded-context foundation

Drakosfire:
## Summary
- Add `statblocks_v1` bounded context (`domain` / `application` / `infrastructure` / `api`) with dependency seams for later PRs.
- Mount `GET /api/internal/dungeonbuddy/v1/statblocks/health` behind router-level DungeonBuddy internal-key auth with a **top-level** typed `ErrorEnvelopeV1` (401/403/503 via exception handlers; misconfigured key env → 503).
- Provide `statblocks_v1.testing.create_test_app()` and a **project-independent** focused test lane.
- AST guards cover the full declared dependency graph (repo-owned packages denied in infrastructure by default); empty protocols are not `@runtime_checkable`; `statblocks_v1/__init__.py` is domain-safe.

## Review follow-ups
1. **Import isolation** — `--confcutdir=tests/statblocks_v1` so `tests/conftest.py` cannot preload `app`.
2. **Dependency isolation** — advertised lane is `./scripts/run_statblocks_v1_tests.sh` (`uv run --isolated --no-project --with pytest/fastapi/pydantic/httpx …`). Ephemeral env (not `.venv`); probe proves openai/firebase/google/fal/sentence_transformers/generationengine are absent.
3. **Import graph** — full layer guards; infrastructure rejects `app`/`routers`/`firestore`/… unless listed in `INFRASTRUCTURE_ADAPTER_EXCEPTIONS`; package root `__init__.py` domain-safe.
4. **Protocols** — placeholders not `@runtime_checkable`.
5. **Typed error envelope** — top-level `{ "error": { ... } }`; OpenAPI 401/403/503; missing config → 503.

## Stack
This is **PR12** of the PR12–PR21 DungeonBuddy statblock v1 series. Successor branches will stack on this head.

## Test plan
- [x] `./scripts/run_statblocks_v1_tests.sh` (19 passed; credentials unset)
- [ ] Spot-check health via isolated app with `DUNGEONBUDDY_INTERNAL_API_KEY` set
- [ ] Confirm legacy `/api/statblockgenerator/*` routes unchanged

### DIFF
--- a/tests/statblocks_v1/test_import_boundaries.py
+++ b/tests/statblocks_v1/test_import_boundaries.py
@@ -20,6 +20,8 @@ from __future__ import annotations
 import ast
 from pathlib import Path
 
+import pytest
+
 REPO_ROOT = Path(__file__).resolve().parents[2]
 PACKAGE_ROOT = REPO_ROOT / "statblocks_v1"
 
@@ -169,22 +171,52 @@ def test_domain_imports_only_stdlib_and_pydantic() -> None:
         _assert_domain_safe_imports(path)
 
 
-def test_application_depends_only_on_domain() -> None:
+def _assert_application_import(name: str, *, path: Path | str = "<synthetic>") -> None:
+    """Enforce application → domain (stdlib/Pydantic allowed; outer layers rejected)."""
     allowed_roots = STDLIB_AND_TYPING_ROOTS | {"pydantic", "statblocks_v1"}
+    root = name.split(".")[0]
+    assert root in allowed_roots, f"{path} imports unexpected root {root!r} ({name})"
+    assert root not in REPO_OWNED_PACKAGE_ROOTS, (
+        f"application must not import repository package {root!r} ({path})"
+    )
+    if name == "statblocks_v1":
+        return
+    layer = _layer_of(name)
+    if layer is not None:
+        assert layer in {"domain", "application"}, (
+            f"application must not import {name!r} ({path})"
+        )
+
+
+def test_application_depends_only_on_domain() -> None:
     for path in _python_files("application"):
         _assert_no_legacy(path)
         for name in _import_names(path):
-            root = name.split(".")[0]
-            assert root in allowed_roots, f"{path} imports unexpected root {root!r} ({name})"
-            assert root not in REPO_OWNED_PACKAGE_ROOTS, (
-                f"application must not import repository package {root!r} ({path})"
-            )
-            if name == "statblocks_v1":
-                continue
-            layer = _layer_of(name)
-            assert layer in {"domain", "application"}, (
-                f"application must not import {name!r} ({path})"
-            )
+            _assert_application_import(name, path=path)
+
+
+def test_application_import_policy_allows_stdlib_pydantic_rejects_outer_layers() -> None:
+    """Stdlib/Pydantic must pass; application → infrastructure/api must fail."""
+    for allowed in (
+        "typing",
+        "datetime",
+        "pydantic",
+        "statblocks_v1",
+        "statblocks_v1.domain",
+        "statblocks_v1.domain.errors",
+        "statblocks_v1.application",
+    ):
+        _assert_application_import(allowed)
+
+    for forbidden in (
+        "statblocks_v1.infrastructure",
+        "statblocks_v1.infrastructure.runtime",
+        "statblocks_v1.api",
+        "statblocks_v1.api.router",
+        "statblocks_v1.testing",
+    ):
+        with pytest.raises(AssertionError):
+            _assert_application_import(forbidden)
 
 
 def test_infrastructure_depends_only_on_domain_and_application() -> None:
--- a/tests/statblocks_v1/test_import_boundaries.py
+++ b/tests/statblocks_v1/test_import_boundaries.py
@@ -1,0 +1,75 @@
+"""Guardrails for statblocks_v1 package dependency direction."""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+PACKAGE_ROOT = Path(__file__).resolve().parents[2] / "statblocks_v1"
+
+
+def _module_imports(path: Path) -> set[str]:
+    tree = ast.parse(path.read_text())
+    imports: set[str] = set()
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Import):
+            for alias in node.names:
+                imports.add(alias.name.split(".")[0])
+        elif isinstance(node, ast.ImportFrom) and node.module:
+            imports.add(node.module.split(".")[0])
+            if node.module.startswith("statblocks_v1"):
+                imports.add(node.module)
+    return imports
+
+
+def _python_files(relative: str) -> list[Path]:
+    return sorted((PACKAGE_ROOT / relative).rglob("*.py"))
+
+
+def test_domain_imports_only_stdlib_and_pydantic() -> None:
+    allowed_roots = {
+        "__future__",
+        "annotations",
+        "typing",
+        "dataclasses",
+        "datetime",
+        "enum",
+        "abc",
+        "collections",
+        "functools",
+        "itertools",
+        "re",
+        "math",
+        "decimal",
+        "uuid",
+        "json",
+        "pydantic",
+        "statblocks_v1",
+    }
+    forbidden_substrings = (
+        "fastapi",
+        "openai",
+        "firebase",
+        "firestore",
+        "statblockgenerator",
+        "cloudflare",
+        "routers",
+    )
+
+    for path in _python_files("domain"):
+        imports = _module_imports(path)
+        for name in imports:
+            assert not any(bad in name for bad in forbidden_substrings), (
+                f"{path} imports forbidden module {name!r}"
+            )
+            root = name.split(".")[0]
+            assert root in allowed_roots or name.startswith("statblocks_v1.domain"), (
+                f"{path} imports unexpected root {root!r}"
+            )
+
+
+def test_api_does_not_import_legacy_statblockgenerator() -> None:
+    for path in _python_files("api"):
+        text = path.read_text()
+        assert "statblockgenerator" not in text
+        assert "StatBlockDetails" not in text
--- a/tests/statblocks_v1/test_import_boundaries.py
+++ b/tests/statblocks_v1/test_import_boundaries.py
@@ -1,4 +1,17 @@
-"""Guardrails for statblocks_v1 package dependency direction."""
+"""Guardrails for statblocks_v1 package dependency direction.
+
+Declared graph:
+
+- domain → standard library and Pydantic only
+- application → domain
+- infrastructure → domain / application (plus external SDKs later)
+- api → domain / application and FastAPI
+
+Narrow exception: ``api`` may import ``routers.internal_auth`` for shared
+internal-key constant names only (not legacy generator packages). The
+foundation prefers local mirrors of those constants so focused tests avoid
+``routers`` package side effects.
+"""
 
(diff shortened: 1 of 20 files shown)


Drakosfire (commented):

The architectural direction is correct: a separate `statblocks_v1` package, narrow production registration, router-level auth, a foundation-only health capability, no reuse of `StatBlockDetails`, and no provider/persistence behavior in this slice. The PR does not yet prove the central acceptance condition: that this bounded context can be developed and tested independently of the legacy application graph.

## Required changes

### 1. Make the focused test lane genuinely isolated

`uv run pytest tests/statblocks_v1/ -q` still discovers the ancestor `tests/conftest.py`, which imports `app` at module import time. That loads the production application and its unrelated dependencies before the focused suite runs. The current `test_create_test_app_does_not_import_production_app` cannot catch this because it records `sys.modules` only after test collection, when `app` may already be loaded.

Required:
- establish a canonical command that cuts off the parent conftest, for example `uv run pytest --confcutdir=tests/statblocks_v1 tests/statblocks_v1 -q`, or relocate/restructure the focused suite so the production conftest is not an ancestor;
- add a reliable guard from a clean interpreter proving `app` is not imported while importing the v1 package, constructing the isolated app, and exercising health;
- update the PR test plan and PR13 predecessor notes with the actual isolated command;
- run that command with external credentials removed.

This is the primary mission of PR12, not a cosmetic test adjustment.

### 2. Enforce the complete dependency direction, not only two fragments

The package documents:

`domain -> stdlib/Pydantic`
`application -> domain`
`infrastructure -> domain/application + SDKs`
`api -> domain/application + FastAPI`

The tests currently enforce only the domain root allow-list and that `api` does not mention the legacy statblock generator. Future PRs can therefore introduce `application -> infrastructure`, `domain -> application`, `api -> infrastructure`, or `infrastructure -> api` without failing this foundation suite.

Required: extend the import-boundary tests to enforce every declared layer direction, including forbidden inward/upward imports and the narrow explicit exception for shared internal-auth code.

### 3. Do not make empty protocols `runtime_checkable`

`GenerationProvider`, `CandidateRepository`, `StatblockRepository`, `RevisionRepository`, and `AssetGateway` have no members but are decorated with `@runtime_checkable`. An empty runtime protocol is structurally satisfied by essentially any object, so runtime checks would provide false confidence rather than a seam.

Required: either give each protocol at least one real capability member, or leave placeholders non-runtime-checkable until later PRs define their methods. `Clock` and `IdAllocator` already have meaningful members and may remain runtime-checkable.

### 4. Return one actual typed v1 HTTP error envelope

The dependency raises `HTTPException(detail={"error": ...})`, so FastAPI emits `{"detail": {"error": ...}}`. That is a framework-shaped payload, not the declared v1 envelope, and it is not represented as a typed response in OpenAPI. The tests currently lock in this accidental nested shape.

Required:
- define the preliminary API error DTO/envelope;
- map auth failures to that top-level response consistently, through an exception handler or explicit response mapping;
- expose the response schema/statuses in OpenAPI;
- use the planned service-unavailable status for missing internal configuration rather than cementing a generic 500 unless the contract document is deliberately amended.

## Non-blocking observations

- Reusing only the constants from `routers.internal_auth` while duplicating comparison behavior is acceptable for this foundation, but a later shared pure auth checker would reduce drift.
- Manual setuptools package enumeration follows the existing repository pattern, though it will become brittle as the bounded context gains subpackages; PR21 is the appropriate cleanup point.

Once the four required items are addressed, this is a strong foundation for PR13.

Drakosfire:
Addressed the blocking review items on this PR:

1. **Isolated test lane** — advertised command is now `uv run pytest --confcutdir=tests/statblocks_v1 tests/statblocks_v1 -q`. Added clean-process and `--trace-config` proofs that `app` / parent `tests/conftest.py` are never loaded; ran with external credentials removed. PR13 predecessor notes updated.
2. **Full layer AST guards** — domain / application / infrastructure / api directions enforced; narrow `routers.internal_auth` exception retained (constants mirrored locally to avoid `routers` package side effects).
3. **Empty protocols** — removed `@runtime_checkable` from placeholder seams; `Clock` / `IdAllocator` unchanged.
4. **Typed error envelope** — `ErrorEnvelopeV1` + `StatblockV1HTTPError` handlers return a top-level `{error:{...}}` body; OpenAPI declares 401/403/503; missing internal key env → **503** `internal_service_misconfigured`.

Ready for re-review.

Drakosfire (commented):
## Re-review verdict: two changes still required before merge

The four issues from the first review are substantively resolved:

- the parent `tests/conftest.py` is cut off and clean-process import isolation is tested;
- layer guards now cover all four package directories;
- empty protocols are no longer runtime-checkable;
- auth errors use a top-level typed envelope with declared 401/403/503 responses.

The foundation is close, but two acceptance conditions are not yet met.

### 1. Make the focused lane dependency-isolated, not only import-isolated

The canonical command is still:

```bash
uv run pytest --confcutdir=tests/statblocks_v1 tests/statblocks_v1 -q
```

Because it runs as a project command from the repository root, `uv run` ensures the project environment contains the dependencies declared by the root `pyproject.toml`. That list still includes OpenAI, Firebase/Firestore, `sentence-transformers`, Fal, and the git-based `generationengine` dependency.

The new subprocess tests prove that the focused suite does not *import* those systems after an environment already exists. They do not prove the PR12 requirement that the lane can start from a clean environment without obtaining the full server dependency/model graph. In particular, a fresh runner may still need network access to the GenerationEngine git repository and may install the large ML stack before pytest begins.

Required:

- provide a project-independent minimal lane, for example a PEP 723 test-runner script or `uv run --no-project --with ...`, containing only the packages needed for this slice (`pytest`, FastAPI, Pydantic, HTTPX/Starlette dependencies as necessary);
- prove that lane from an empty disposable environment/cache or CI job;
- ensure the lane does not resolve or install `generationengine`, OpenAI, Firebase, Cloudflare, or `sentence-transformers`;
- update the PR description, PR12 handoff, PR13 predecessor notes, and isolation test to advertise and exercise that command.

The handoff explicitly requires a focused lane that does not require the full server’s external dependencies or model downloads; `--confcutdir` solves collection coupling but not installation coupling.

### 2. Close the remaining infrastructure-boundary loophole

`test_infrastructure_depends_only_on_domain_and_application` currently forbids `fastapi` and `statblockgenerator`, plus `statblocks_v1.api`, but otherwise accepts any import root. It would therefore allow future infrastructure code to import repository-owned legacy modules such as:

```text
app
routers
firestore.firebase_config
cloudflare
ruleslawyer
cardgenerator
```

That would pass the guard while recreating exactly the import-time and cross-product coupling this bounded context is intended to prevent.

Required:

- distinguish external SDK roots from repository-owned modules and reject imports from other repository packages by default;
- use an explicit, narrow allow-list only when a later adapter deliberately reuses a proven internal module;
- guard `statblocks_v1/__init__.py` as a domain-safe package root, since domain/application are currently allowed to import `statblocks_v1`; otherwise a future root re-export of API or infrastructure code would create a transitive boundary violation that these tests would miss.

A practical guard can derive repository-owned roots from top-level Python files/directories instead of maintaining a long hard-coded list.

Once these two points are addressed, I see no remaining PR12-level blocker. The typed error handling, route registration, scope, successor notes, and protocol treatment are otherwise ready for PR13.

Drakosfire:
Addressed the latest two blocking items:

1. **Dependency-isolated lane** — `./scripts/run_statblocks_v1_tests.sh` runs `uv run --isolated --no-project` with only `--with pytest/fastapi/pydantic/httpx`, plus `--confcutdir=tests/statblocks_v1`. This uses an ephemeral env (not the project `.venv`) so a clean machine never syncs OpenAI/Firebase/sentence-transformers/generationengine. A probe test asserts those modules are absent. PR13 notes updated.

2. **Infrastructure / package-root guards** — repository-owned packages (`app`, `routers`, `firestore`, `cloudflare`, `ruleslawyer`, `cardgenerator`, …) are rejected by default; adapters opt in via `INFRASTRUCTURE_ADAPTER_EXCEPTIONS`. `statblocks_v1/__init__.py` is guarded as domain-safe.

Ready for re-review.

tests/statblocks_v1/test_import_boundaries.py
```diff
@@ -0,0 +1,265 @@
...
+            assert root in allowed_roots, f"{path} imports unexpected root {root!r} ({name})"
+            assert root not in REPO_OWNED_PACKAGE_ROOTS, (
+                f"application must not import repository package {root!r} ({path})"
+            )
+            if name == "statblocks_v1":
+                continue
+            layer = _layer_of(name)
+            assert layer in {"domain", "application"}, (
+                f"application must not import {name!r} ({path})"
+            )
```
Drakosfire: This assertion also runs for permitted stdlib/Pydantic imports, where `_layer_of(name)` returns `None`. As written, the first future application module that imports `typing`, `datetime`, or Pydantic will fail the boundary suite. Guard the assertion with `if layer is not None`, then add a test case proving allowed external roots pass while upward v1-layer imports still fail.

Drakosfire (commented):
## Re-review verdict: one required fix remains

The two previous blockers are now genuinely addressed:

- `scripts/run_statblocks_v1_tests.sh` is executable and runs through an ephemeral `uv run --isolated --no-project` environment with only the focused test dependencies.
- Infrastructure imports are denied against dynamically discovered repository-owned packages unless a file-specific adapter exception explicitly permits them.

The earlier isolation, protocol, typed-error, and OpenAPI issues also remain fixed.

### Required: correct the application-layer import guard

`test_application_depends_only_on_domain()` first permits stdlib and Pydantic roots, but then unconditionally requires every non-bare import to resolve to a `statblocks_v1` layer:

```python
layer = _layer_of(name)
assert layer in {"domain", "application"}
```

For a valid application import such as `from typing import Protocol`, `from datetime import datetime`, or `from pydantic import BaseModel`, `_layer_of(name)` returns `None`, so this guard fails despite those roots being explicitly allowed above.

Make this consistent with the API and infrastructure guards:

```python
layer = _layer_of(name)
if layer is not None:
    assert layer in {"domain", "application"}
```

Add a small fixture/source case or helper-level test proving the application layer may import stdlib and Pydantic while still rejecting `statblocks_v1.infrastructure` and `statblocks_v1.api`.

After that correction, I see no remaining PR12-level blocker.

Drakosfire:
Fixed the remaining application-layer import guard bug: stdlib/Pydantic imports no longer fail when `_layer_of()` returns `None`. Added `test_application_import_policy_allows_stdlib_pydantic_rejects_outer_layers` proving typing/datetime/pydantic pass while `application → infrastructure` and `application → api` remain rejected.

`./scripts/run_statblocks_v1_tests.sh` → 20 passed.

Drakosfire (commented):
## Re-review verdict: merge-ready

The latest commit fixes the final application-layer import-guard defect correctly.

- `_assert_application_import()` now permits stdlib and Pydantic imports without applying a `statblocks_v1` layer assertion.
- Imports from `statblocks_v1.domain` and `statblocks_v1.application` remain allowed.
- Imports from infrastructure, API, and sibling modules such as `statblocks_v1.testing` remain rejected.
- The new policy test directly covers representative accepted and forbidden imports, preventing regression.

The earlier isolation, dependency-lane, typed-error, protocol, and repository-boundary concerns remain addressed. I found no remaining PR12-scoped blocker.

GitHub does not allow this account to formally approve its own pull request, so this comment records the equivalent review verdict: **PR12 is ready to merge.**

No CI status checks are currently attached to the head commit; the PR reports `./scripts/run_statblocks_v1_tests.sh` passing 19 tests.

PR #13, opened 2026-07-18
feat(statblocks_v1): PR13 contract models and fixtures

Drakosfire:
## Summary
- Add executable Pydantic v2 `StatblockDefinitionV1` models (primitives, profiles, rule elements) with `extra=forbid` and snake_case JSON.
- Add canonical + OpenAI-strict schema artifacts, representative valid/invalid fixtures, and field-disposition mapping from legacy `StatBlockDetails`.
- Focused domain tests for parse/round-trip (exact fixture JSON parity), discriminators, and schema snapshots.

## Review follow-ups (this push)
1. **Rebased onto `main`** (merged PR12 foundation); PR base retargeted to `main`.
2. **Schema compiler** — context-aware metadata stripping (preserves `ArmorClassProfile.default`, `StatblockFlavorText.description`); `oneOf` → `anyOf`; `Usage.recharge_range` is `{minimum,maximum}` object (no `prefixItems`); recursive unsupported-keyword + property-path parity tests.
3. **Proficiency authority** — authored only on `challenge.proficiency_bonus` (removed from `ProficiencyProfile`).
4. **Field disposition** — legendary intro, lair name/both descriptions, spell name/level/school.
5. **Fixture round trips** — `model_dump(mode=\"json\") == original fixture JSON`.

## Stack
**PR13** of PR12–PR21. Base: `main` (PR12 merged).

## Test plan
- [x] `./scripts/run_statblocks_v1_tests.sh` (44 passed; isolated runner, credentials unset)
- [ ] Review schema snapshot diffs for unintended contract drift
- [ ] Spot-check one advanced fixture (legendary / mythic / spellcaster)


Drakosfire (commented):

The domain split, finite mechanic/effect unions, fixture breadth, strict `extra="forbid"` base, and successor handoff are strong. The implementation is close to the approved canonical shape. The blockers are at the places where this PR becomes contract authority: branch ancestry, provider-schema equivalence, single-source mechanics, and migration accounting.

## Required changes

### 1. Rebase this PR onto the merged PR12 foundation and retarget it to `main`

The PR is still based on `feat/statblocks-v1-pr12-foundation` at the early PR12 SHA, rather than the final merged PR12 state. GitHub currently reports the PR as non-mergeable, and the PR test plan still advertises `uv run pytest tests/statblocks_v1/ -q` instead of PR12's project-independent `./scripts/run_statblocks_v1_tests.sh` lane.

Required:
- rebase the PR13 branch onto current `main`;
- retarget the PR base to `main`;
- resolve the final PR12 test runner/import-boundary changes rather than carrying the early foundation snapshot;
- run and record `./scripts/run_statblocks_v1_tests.sh` after the rebase;
- update the PR body and PR14 predecessor notes if paths or test counts change.

### 2. Fix the OpenAI strict-schema compiler; the checked artifact is neither semantically equivalent nor provider-safe

There are three concrete failures.

#### 2a. Metadata stripping deletes real contract properties

`_transform_for_openai` removes every dictionary key named `default` or `description` without knowing whether it is looking at schema metadata or a `properties` mapping. Consequently, legitimate fields disappear from the provider contract:

- `ArmorClassProfile.default` exists in the Pydantic model but is absent from the strict artifact;
- `StatblockFlavorText.description` exists in the model but the strict artifact contains only `summary`.

A model generated against that schema cannot round-trip into the canonical Pydantic definition. This directly violates the requirement that provider transformation preserve semantic fields.

Required: make transformation context-aware. Strip schema metadata only from schema nodes; never filter user property names from a `properties` map. Add regression tests for at least `ArmorClassProfile.default`, `LairProfile.description`, and `StatblockFlavorText.description`.

#### 2b. Discriminated unions remain `oneOf`

The strict artifact contains `oneOf` for `Effect`/`Mechanic` discriminated unions. OpenAI Structured Outputs documents `anyOf` as the supported union form; `oneOf` is not in the supported subset. Removing the `discriminator` keyword does not make the remaining `oneOf` valid.

Required: compile provider unions to supported `anyOf` branches while preserving each literal `kind` discriminator field and branch object closure. Add a test that recursively rejects `oneOf`, `allOf`, and other unsupported composition keywords.

#### 2c. `Usage.recharge_range` emits `prefixItems`

The tuple model generates a fixed-tuple schema using `prefixItems`. That keyword is not part of the documented Structured Outputs subset.

Required: replace the tuple with a provider-safe explicit object such as `RechargeRange(minimum, maximum)`, or perform a semantics-preserving provider transformation with matching parsing. An explicit object is clearer canonical JSON and avoids positional ambiguity.

The current schema test only checks that six keywords were removed and every object is closed; it therefore approves an artifact that has lost fields and still contains unsupported constructs. Add a pure subset validator and a semantic property-path parity test between canonical and strict schemas. A gated live provider smoke belongs later, but PR13 must not publish an artifact already known to be invalid.

### 3. Eliminate the duplicated canonical proficiency bonus

The definition currently authors `proficiency_bonus` in both `ProficiencyProfile` and `ChallengeProfile`, and the field-disposition table explicitly maps the one legacy value into both places.

That creates two canonical values for one deterministic mechanic and leaves PR14 to police drift after the schema has already admitted it. The contract direction is one mechanics definition without redundant co-authority.

Required: choose one canonical source. The cleanest current shape is `challenge.proficiency_bonus`; saving-throw and skill derivation can reference that value. Remove `proficiencies.proficiency_bonus` from models, fixtures, artifacts, and the field-disposition table, or amend the authoritative design with a compelling distinct semantic reason for both fields and explicit override ownership.

### 4. Complete the legacy field-disposition artifact

The table covers the top-level `StatBlockDetails` fields and the shallow `Action` shape, but it does not account for all nested legacy fields. Missing examples include:

- `LegendaryActionsBlock.description`;
- `LairActionsBlock.lair_name`;
- `LairActionsBlock.lair_description`;
- `LairActionsBlock.description`;
- `Spell.name`;
- individual `Spell.level`;
- `Spell.school`.

The acceptance criterion is that all legacy fields receive an explicit destination or removal decision. Add those rows, even where the decision is “removed,” “derived,” or deferred to a future canonical spell locator. Do not silently lose data during the architectural reset.

### 5. Strengthen the fixture round-trip assertion

`test_valid_fixture_parses_and_round_trips` currently proves only that a dumped model reparses to an equal model. It does not prove that the original complete JSON fixture survives parsing unchanged.

Required: for complete PR13 fixtures, assert `definition.model_dump(mode="json") == payload` as well as reparsing. PR14 may later introduce documented canonical normalization; PR13 should expose any field dropping or default insertion now.

## Non-blocking observations

- Reserving duplicate-key, dangling-reference, armor-cardinality, and section/activation checks for PR14 matches the handoff.
- `PassiveMechanic` and `CompositeMechanic` correctly follow the authoritative design and the PR14 handoff records the intentional divergence from the earlier movement/utility wording.
- The fixture set is usefully advanced rather than attack-only.

Once these items are addressed, the model layer will be a credible contract authority rather than only a comprehensive Pydantic draft.

Drakosfire:
Addressed the consolidated blocking review:

1. Rebased onto `main` and retargeted the PR base; test plan now uses `./scripts/run_statblocks_v1_tests.sh`.
2. Schema compiler: context-aware metadata stripping; `oneOf`→`anyOf`; `RechargeRange` object; recursive keyword + property-path parity tests. `default`/`description` property names preserved.
3. `challenge.proficiency_bonus` is the sole authored proficiency bonus.
4. Field disposition completed for legendary intro, lair name/descriptions, and spell name/level/school.
5. Fixtures assert exact `dump == original` JSON parity.

`./scripts/run_statblocks_v1_tests.sh` → 44 passed.

Drakosfire (commented):
## Re-review verdict: changes still required before merge

The original five blockers are materially addressed:

- the PR is rebased and retargeted to `main` and is mergeable;
- the isolated runner is restored and pins Pydantic at `2.7.4` for reproducible schema snapshots;
- provider property names such as `default` and `description` survive compilation;
- discriminated unions compile to `anyOf`, and recharge range is now an explicit object;
- proficiency bonus has one authored authority;
- the migration table and exact fixture round trips are substantially stronger.

Two contract-level issues remain.

### 1. The provider-schema compiler must fail closed instead of silently deleting unsupported constraints

`_transform_schema_node()` currently handles `oneOf` with a semantics-preserving rewrite, but for every other entry in `_UNSUPPORTED_OPENAI_KEYWORDS` it simply executes `continue`:

```python
if key in _UNSUPPORTED_OPENAI_KEYWORDS:
    # Drop remaining unsupported constructs; model surface should avoid them.
    continue
```

This means a future canonical-model change that introduces `allOf`, `not`, `if/then/else`, `dependentRequired`, or another unsupported construct will produce a weaker provider schema while the current tests remain green. Property-path parity cannot detect lost constraints, and the recursive keyword test rewards deletion rather than proving equivalence.

That violates the contract rule that provider compilation may rewrite syntax but may not create a semantically different payload.

Required:

- retain explicit lossless rewrites such as `oneOf` → `anyOf`;
- raise a typed compilation error on every unsupported keyword without a documented lossless rewrite;
- add a synthetic regression test proving unsupported canonical input fails compilation rather than disappearing;
- ideally assert that the current canonical schema contains no unsupported constructs except the explicitly rewritten ones.

Metadata removal (`title`, schema-level `default`, etc.) is separate and may remain intentional. OpenAI currently documents `anyOf`, definitions, pattern/numeric/array constraints, required fields, and closed objects as supported, while unsupported composition schemas cause strict requests to fail.

### 2. Give spell level one unambiguous canonical authority and correct the innate fixture

The approved design places `level` on `SpellGroup`; this PR adds another optional `SpellRef.level`. The current fixtures reinforce that `SpellGroup.level` is the active authority: both spellcaster fixtures set group levels while every spell-level field is `null`.

As written, the schema can now say a spell belongs to a level-3 group while `SpellRef.level` says 5. That repeats the same co-authority problem just removed from proficiency bonus. The field-disposition table claims legacy `Spell.level` maps directly to `SpellRef.level`, but the representative fixtures instead map it through `SpellGroup.level`.

Required: choose and document one shape.

Preferred current-contract option:

- remove `SpellRef.level`;
- map legacy `Spell.level` by grouping spells under `SpellGroup.level`;
- retain `SpellRef.school`, which is not duplicated elsewhere.

If group level and spell level are intended to represent genuinely different concepts, rename them to those concepts (for example intrinsic spell level versus slot/casting level), document the distinction, populate fixtures that demonstrate it, and add the corresponding PR14 consistency policy. Do not leave two fields both named `level` with implicit semantics.

Also correct `innate_spellcaster.json`: an innate `3/day` spell should use `usage.kind = per_day`, `usage.uses = 3`, and `slots = null`. Its current `slots = 3` models slot economy inside an innate caster and makes a required “valid fixture” semantically misleading. Record the slot-versus-innate group rule in the PR14 predecessor notes; the full validator can still land in PR14.

## Non-blocking observation

The compiler strips schema descriptions even though current OpenAI Structured Outputs supports descriptions. Keeping them could improve model generation, but this is a token-budget/product choice rather than a merge blocker.

After the two issues above are corrected, I see no remaining PR13-level blocker.

Drakosfire (commented):
## Review verdict: ready to merge

Re-reviewed current head `9b4d628b571449a8dc44af7fa816a23a0de91c2a` against the PR13 handoff, merged PR12 foundation, and the prior blocking reviews.

The final two blockers are resolved:

- The OpenAI strict-schema compiler now fails closed on unsupported constructs instead of silently deleting constraints. It retains explicit semantics-preserving rewrites for the current schema (`oneOf` discriminated unions to `anyOf`, and Pydantic's single-branch `allOf` wrapper), with regression tests for unsupported `not`, conditional schemas, and multi-branch `allOf`.
- Spell level now has one authored authority at `SpellGroup.level`; `SpellRef.level` is removed from the model, artifacts, fixtures, design text, and field-disposition mapping. The innate fixture correctly represents three daily uses through `Usage(kind="per_day", uses=3)` with `slots=null`.

The earlier changes also remain intact: the PR targets `main`, the provider and canonical schemas retain matching property paths, proficiency bonus is authored only once, fixture dumps exactly match source JSON, and the legacy field mapping covers the previously omitted nested fields.

No remaining PR13-level blocker. GitHub currently reports the PR mergeable. There are no attached CI status checks; the PR reports the isolated test lane passing.

PR #14, opened 2026-07-18
feat(statblocks_v1): PR14 canonicalization, validation, and digests

Drakosfire:
## Summary
- Pure-domain trust core on merged PR13/`main`: sole `challenge.proficiency_bonus`, typed `RechargeRange`, SpellGroup-level authority, `spell_slots` usage.
- `is_persistence_ready` is mode-safe; digests accept only `StatblockDefinitionV1` (canonicalization is internal).
- Completes usage, spellcasting, resource affordability, attack, and proficiency derivation validation; nested effect issues report real collection paths.
- `SkillBonus.ability` is required typed authority; `standard`/`expertise` values are checked; duplicate save abilities and NFC+casefold skill names are rejected.

## Test plan
- [x] `./scripts/run_statblocks_v1_tests.sh` (91 passed)
- [x] Valid fixtures persistence-ready under persistence mode
- [x] Candidate receipt never claims persistence readiness
- [x] Digest API accepts only `StatblockDefinitionV1`
- [x] Nested unknown-element path uses `mechanic.hit_effects[i]`
- [x] Incorrect standard/expertise save/skill values blocked; explicit overrides accepted
- [x] Duplicate save abilities and NFC/casefold-equivalent skill names rejected


Drakosfire (commented):

The pure-domain split, stable issue-code direction, versioned canonicalizer, bounded rules-text contradiction checks, and fixture-driven intent are all strong. The blockers are at the exact points where PR14 becomes the trust authority consumed by immutable persistence.

## Required changes

### 1. Rebase onto the merged PR13 head and retarget to `main`

This PR is still based on the pre-review PR13 SHA `1c062b1...`, not the merged PR13 contract, and GitHub currently reports it as non-mergeable. The branch therefore carries contract shapes that PR13 deliberately removed or replaced.

Concrete breakages after a correct rebase:

- `_validate_profiles` reads `definition.proficiencies.proficiency_bonus`, but merged PR13 removed that field and made `challenge.proficiency_bonus` the sole authority.
- recharge validation indexes `recharge[0]` / `recharge[1]`, but merged PR13 replaced the tuple with `RechargeRange.minimum` / `.maximum`.
- the branch contains the pre-fix fixtures, schema artifacts, field-disposition table, and spell model, so merging it as stacked would reintroduce resolved PR13 defects.

Required:

- rebase this branch onto current `main` and retarget the PR to `main`;
- preserve all final PR13 decisions during conflict resolution;
- update validation for `RechargeRange` and the sole proficiency authority;
- regenerate digest snapshots and the PR15 predecessor notes from the rebased contract;
- run and advertise `./scripts/run_statblocks_v1_tests.sh`, not the project-environment `uv run pytest ...` command.

### 2. Make `is_persistence_ready` mode-safe

`ValidationReceiptV1.is_persistence_ready` currently returns true whenever the receipt has no error-severity issues, regardless of validation mode.

That is incorrect because `generation_candidate` and `editor_preview` deliberately downgrade high-confidence rules-text contradictions to warnings. A candidate-mode receipt for a definition that would fail persistence therefore currently reports `is_persistence_ready == True`.

Required:

- either require `mode is ValidationMode.persistence` as part of `is_persistence_ready`, or remove the property and expose a persistence-only result explicitly;
- add a regression test proving that a candidate-mode contradiction cannot be treated as persistence-ready, while a clean persistence-mode receipt can.

PR15 must never be able to persist a receipt merely because a weaker mode contained no errors.

### 3. Complete the required usage, spell-group, passive-Perception, and attack-target validation

The handoff requires coherent usage/recharge fields, coherent spellcasting groups, passive Perception agreement or override semantics, and valid attack target/range behavior. The current implementation only checks whether `recharge_range` is present for `kind == recharge`.

It still accepts examples such as:

- `at_will` usage with `uses` or `resource_key` populated;
- `resource` usage without `resource_key`;
- `per_day` usage without an authored use count where the group is not slot-based;
- innate spell groups represented with `slots` rather than `Usage.uses`;
- slot-bearing cantrip groups or other contradictory `slots` / `level` / casting-mode combinations;
- a typed Perception skill with a passive-Perception value that does not equal `10 + skill.value` (the current code skips validation whenever Perception exists);
- attack targets whose `kind`, `count`, `area`, reach, and range are mutually incoherent.

Required:

- define and test one explicit kind-specific `Usage` policy;
- define and test slot-based versus innate/per-use `SpellGroup` policy using the merged PR13 fixtures;
- validate passive Perception against the typed Perception bonus when present, and against Wisdom otherwise, with explicit override semantics if divergence is allowed;
- add bounded target/reach/range coherence checks for attacks.

### 4. Emit real, UI-addressable paths for nested effects

Reference validation flattens attack, save, phase, passive, and composite effects, then reports every nested issue at:

`rule_elements[i].mechanic.effects[j]`

That path does not exist for attacks (`hit_effects` / `miss_effects`) or saves (`failure_effects` / `success_effects`). DungeonBuddy cannot focus or annotate the offending field from these receipts, and the path may identify the wrong effect after flattening two arrays.

Required:

- have effect traversal retain the actual collection name and original index;
- emit paths such as `.hit_effects[0]`, `.miss_effects[0]`, `.failure_effects[0]`, etc.;
- add tests for unknown movement/resource/phase references in multiple effect collections.

Stable issue codes are not enough if their locations are fictional.

### 5. Do not let callers mint canonical-looking digests from arbitrary strings or bytes

`compute_definition_digest` publicly accepts `StatblockDefinitionV1 | str | bytes` and hashes raw string/byte inputs while stating that callers are responsible for canonicality.

That bypass is incompatible with this module's role as the trust core. Any caller can hash noncanonical JSON and receive a normal-looking `sha256:<hex>` value, and PR15 is explicitly instructed to persist the digest as durable truth.

Required:

- make the public digest function accept `StatblockDefinitionV1` only; or
- introduce a branded/typed canonical payload produced only by `canonicalize_definition`, and accept that type rather than arbitrary `str` / `bytes`.

Add a test demonstrating that whitespace-reordered or otherwise noncanonical raw JSON cannot bypass canonicalization through the public API.

### 6. Replace tautological determinism tests with the handoff's required equivalence tests

`canonical_definition_bytes(definition) == canonical_definition_bytes(definition)` only proves that one immediate function call equals itself. The suite does not currently prove several explicit PR14 requirements.

Required tests:

- two definitions differing only in set-like order/duplicates produce byte-identical canonical JSON and the same digest;
- NFC and canonically equivalent decomposed Unicode normalize identically;
- presentation/sequence order remains digest-significant;
- repeated validation produces equivalent receipts except caller-supplied `validated_at`;
- invalid nested references emit the expected stable code and exact real field path;
- digest snapshots are regenerated after the PR13 rebase.

## Non-blocking observations

- Computing a digest for an invalid definition is useful for diagnostics and is acceptable as long as persistence readiness remains explicit.
- Warnings may remain persistence-ready after persistence-mode validation; the problem is weaker modes claiming that readiness.
- The bounded rules-text parser is appropriately conservative for this PR and does not need to become a general rules-language parser.

After the rebase and these trust-boundary corrections, the implementation should be in good shape for PR15 to consume without duplicating or second-guessing validation behavior.

Drakosfire:
Addressed the blocking review:

1. Rebased onto merged PR13/`main` and retargeted the PR base to `main`.
2. Removed checks against deleted `proficiencies.proficiency_bonus`; recharge validation uses typed `RechargeRange.minimum`/`maximum`.
3. `is_persistence_ready` is true only for persistence-mode receipts with no errors.
4. Completed usage/spellcasting/perception/attack coherence validation.
5. Nested effect findings preserve real collection paths (`hit_effects` / `failure_effects` / etc.).
6. Digest API accepts only `StatblockDefinitionV1` or branded `CanonicalDefinitionJSON`.
7. Replaced tautological byte-identity tests with set-order, NFC/NFD, receipt-time, nested-path, and refreshed digest snapshot coverage.

Focused lane: 79 passed via `./scripts/run_statblocks_v1_tests.sh`.

Drakosfire (commented):
## Re-review verdict: four required fixes remain

The new commit materially resolves the previous review: PR14 is rebased onto `main`, GitHub reports it mergeable, validation uses the sole `challenge.proficiency_bonus`, `RechargeRange` is handled as an object, persistence readiness is mode-safe, nested effect paths preserve their real collection names, and the determinism tests are no longer tautological.

Four trust-core issues still block merge.

### 1. `CanonicalDefinitionJSON` does not actually prevent raw digest bypasses

`CanonicalDefinitionJSON` is a public subclass of `str`. `@final` only affects static type checking; any caller can construct it directly:

```python
compute_definition_digest(CanonicalDefinitionJSON('{"z":1,"a":2}'))
```

That value passes the runtime `isinstance` check and receives a normal-looking `sha256:` digest without ever going through `canonicalize_definition`. The new test rejects plain `str` and `bytes`, but does not test direct construction of the branded class.

Required: make the public digest API accept only `StatblockDefinitionV1`, or use a runtime provenance mechanism that cannot be obtained by calling a public string constructor. The simplest safe API is `compute_definition_digest(definition: StatblockDefinitionV1)`; persistence can store the canonical text returned alongside that digest rather than re-digest arbitrary text.

### 2. Nested spell-group resource references are never validated

`_validate_references` checks `element.usage.resource_key` and element costs, but does not walk `SpellcastingMechanic.groups[].usage.resource_key`. Meanwhile charge-based spell groups are required to use `UsageKind.resource`.

A charge group can therefore reference a nonexistent resource pool and still be persistence-ready. Required: validate every nested group usage resource key against `definition.resources`, with an exact path such as:

```text
rule_elements[i].mechanic.groups[j].usage.resource_key
```

Add a mutation test for an unknown charge-pool reference.

### 3. Attack range still has two unchecked authorities

`AttackMechanic` owns `reach` / `range`, while `TargetProfile` independently permits `range`. The new attack validator checks melee-versus-ranged presence and target count/area, but never checks `mechanic.target.range`.

A definition can therefore carry `mechanic.range = 30/120 feet` and `mechanic.target.range = 5/10 feet` and pass validation. Required: define one authority for attack distance. Prefer requiring `target.range` to be absent for `AttackMechanic`, since attack reach/range already owns that fact; otherwise require exact equality and document the distinct semantics.

### 4. Slot-backed spellcasting is being mislabeled as `at_will`

The updated prepared-spellcaster fixture now represents a level-3 group with three slots as:

```json
{"usage":{"kind":"at_will"},"level":3,"slots":3}
```

and validation requires all leveled prepared/known groups to use `at_will`. In ordinary statblock language, `at_will` means no expendable limit; three spell slots are explicitly limited. This makes the canonical typed statement internally misleading even if downstream code learns to special-case it.

The authoritative design says `SpellGroup.usage` supports normal slots, innate at-will, per-day, and charge-based casting, but does not define `at_will` as the slot-backed mode. Required: choose and document an unambiguous slot-backed representation. A reasonable v1 policy is `per_day` (or another explicit refresh mode) plus `slots`, with `slots` supplying the count and no duplicate `uses`; alternatively add an explicit slot-backed usage kind. Do not overload `at_will` to mean "available subject to slots."

## Verified fixes

- Base is now `main`; current head is mergeable.
- `ValidationReceiptV1.is_persistence_ready` requires persistence mode.
- Sole proficiency authority and typed recharge handling are correct.
- Passive Perception now checks an authored Perception bonus.
- Nested effect paths are preserved.
- Set-like ordering, Unicode normalization, presentation ordering, repeated receipts, and refreshed digest snapshots are covered.

After these four corrections, I see no remaining issue from the prior six-item review.

Drakosfire:
Follow-up trust fixes:

1. `compute_definition_digest` now accepts only `StatblockDefinitionV1` — forged `CanonicalDefinitionJSON('...')` is rejected.
2. `SpellGroup.usage.resource_key` is validated against resource pools (charges cannot point at missing pools).
3. `AttackMechanic.target.range` is rejected; distance authority is `mechanic.reach` / `mechanic.range` only.
4. Added `UsageKind.spell_slots` for leveled prepared/known groups; cantrips remain `at_will`. Fixture + schema artifacts updated.

Focused lane: 81 passed.

Drakosfire (commented):
## Re-review verdict: three changes remain before merge

The new commit correctly resolves the four prior blockers:

- `compute_definition_digest` now accepts only `StatblockDefinitionV1`, so forged canonical-string wrappers cannot enter the digest path;
- nested `SpellGroup.usage.resource_key` values are reference-checked;
- attacks reject `target.range`, leaving reach/range authority on `AttackMechanic`;
- slot-backed prepared/known groups now use an explicit `spell_slots` usage kind rather than overloading `at_will`.

The branch is also based on `main` and currently mergeable. Three trust-core gaps remain.

### 1. Restrict `spell_slots` to the context it was introduced for

`UsageKind.spell_slots` is now part of the global `Usage` enum, and `_validate_usage` accepts it wherever a `Usage` appears. That means both of these currently pass generic usage validation:

- an ordinary attack or trait whose top-level `rule_element.usage.kind` is `spell_slots`;
- an innate spell group using `spell_slots` with no slots.

The contract text says this kind marks **slot-backed prepared/known spell groups**. Validation should enforce that scope rather than treating it as a generally valid usage mode.

Required:

- reject `spell_slots` on top-level rule-element usage;
- reject it for innate and charge-based spell groups;
- allow it only for prepared/known groups with `level > 0` and a positive `slots` value;
- add regression tests for both invalid contexts and the valid prepared/known case.

A contextual flag or separate spell-group usage validator is preferable to leaving the generic `_validate_usage` API unable to distinguish where it is being used.

### 2. Make legendary-action usage and cost pools agree

Legendary validation currently checks:

- usage kind is `resource`;
- at least one cost exists;
- a cost points to some existing resource pool.

Reference validation separately proves that `usage.resource_key` and each cost key exist. It does not prove they identify the same resource.

A definition can therefore say:

```text
usage.resource_key = legendary_actions
costs[0].resource_key = mythic_points
```

with both pools present and still be persistence-ready. That leaves two contradictory authorities for what the legendary action consumes.

Required:

- require at least one legendary-action cost to use `element.usage.resource_key`;
- if multi-pool legendary costs are not intentionally supported, require every cost to use that same key;
- add a mutation test with two valid pools and mismatched usage/cost keys.

### 3. Validate normal/long range ordering

`RangeProfile` structurally permits any nonnegative `normal` and `long` distances, and attack validation currently checks only whether ranged attacks have a range. A ranged attack with normal range 120 feet and long range 30 feet is therefore persistence-ready.

Required:

- when `long` is present, require `long.value >= normal.value`;
- report a stable, field-addressable validation issue, preferably at `.mechanic.range.long`;
- add a focused mutation test.

This can live in the Pydantic model or the PR14 validator, but because PR14 owns persistence readiness and stable issue codes, domain validation is the more useful location for DungeonBuddy feedback.

## Non-blocking

The `spell_slots` addition is otherwise integrated consistently into the design text, field-disposition mapping, schema artifacts, fixture, digest snapshot, and PR15 handoff. The isolated suite reported in the PR remains useful, though no CI status checks are attached to this head.

Drakosfire:
Three remaining trust fixes:

1. `spell_slots` is rejected everywhere except leveled prepared/known `SpellGroup`s (ordinary rule-element usage and innate groups fail).
2. Legendary actions require `usage.resource_key` to match every `costs[].resource_key` (`LEGENDARY_RESOURCE_MISMATCH`).
3. `RangeProfile` rejects long < normal; attack validation also emits `ATTACK_RANGE_ORDER_INCOHERENT`.

Focused lane: 85 passed.

Drakosfire (commented):
## Re-review verdict: one required change remains

The latest commit correctly resolves all three findings from the previous review:

- `spell_slots` is rejected at rule-element level and outside leveled prepared/known spell groups;
- legendary usage and costs must name the same existing resource pool;
- `RangeProfile` and defensive attack validation reject inverted or unit-mismatched normal/long ranges.

The branch remains mergeable and the new regression tests exercise each correction.

### Required: reject resource costs that can never be paid

`ResourcePool.maximum` is the maximum available amount, while every `ResourceCost.amount` is authored independently. The current validator confirms that cost keys resolve—and, for legendary actions, that they match `usage.resource_key`—but never checks cost amount against the referenced pool's maximum.

A definition such as:

```python
ResourcePool(key="legendary_actions", maximum=3, ...)
ResourceCost(resource_key="legendary_actions", amount=4)
```

currently passes persistence validation even though that action can never be used. The same problem applies to any rule element with resource costs, not only legendary actions. Multiple cost entries against the same pool can also exceed the maximum in aggregate.

Required:

- validate resource costs against the referenced `ResourcePool.maximum`;
- for repeated costs against one pool, either reject duplicate entries or compare their summed amount with the pool maximum;
- emit a stable, UI-addressable issue code/path such as `RESOURCE_COST_EXCEEDS_POOL` at the offending `costs[i].amount` or cost collection;
- add regression coverage for an individually excessive cost and, if duplicate same-pool costs remain legal, an excessive aggregate.

This belongs in PR14 because PR15 is instructed to trust `is_persistence_ready` rather than recreate mechanics validation.

### Cleanup before merge

The PR description is stale: it still reports 79 tests and says the digest API accepts `CanonicalDefinitionJSON`, although the current API intentionally accepts only `StatblockDefinitionV1`. Please refresh the summary/test plan after the final test run.

There are no GitHub status checks attached to this head, so I could not independently confirm the reported isolated-suite result.

Drakosfire:
Resource-cost affordability blocker fixed:

- `RESOURCE_COST_EXCEEDS_POOL` at `rule_elements[i].costs[j].amount` when a cost (or same-pool aggregate) exceeds `ResourcePool.maximum`
- `RESOURCE_COST_DUPLICATE_POOL` at `rule_elements[i].costs[j].resource_key` when the same pool appears more than once in `costs[]`
- Regression tests for amount=4 vs max=3 and aggregate 2+2 vs max=3

Focused lane: all green.

Drakosfire (commented):
## Review verdict: one trust-core blocker remains

The latest commit correctly closes the previously reported resource-affordability defect:

- each `ResourceCost.amount` is checked against the referenced `ResourcePool.maximum`;
- duplicate same-pool costs are rejected rather than ambiguously summed;
- aggregate overflow is detected;
- stable issue codes and exact UI-addressable paths are covered by focused tests;
- the PR15 handoff includes the new issue codes.

That targeted fix is good.

### Required: validate proficiency derivations and duplicate authorities

The contract explicitly makes `SavingThrowBonus.derivation` and `SkillBonus.derivation` authoritative typed semantics (`standard | expertise | explicit_override`) and states that DungeonMindServer verifies standard derivations. The trust core currently does not inspect either derivation field at all.

As a result, persistence validation currently accepts examples such as:

- Strength 18, proficiency bonus +3, but a saving throw declared `standard` with value `+2`;
- a canonical skill declared `standard` or `expertise` with a value that does not match its ability modifier and proficiency bonus;
- two authored saving-throw entries for the same ability;
- duplicate skill entries differing only by case, such as `Perception` and `perception`.

The duplicate case is especially problematic because consumers receive multiple co-authoritative values, and passive-Perception validation currently selects the first matching entry.

Required before merge:

1. Validate saving throws deterministically:
   - `standard = ability modifier + challenge.proficiency_bonus`;
   - `expertise = ability modifier + 2 × challenge.proficiency_bonus`;
   - `explicit_override` is preserved without forcing the standard arithmetic.
2. Give skills enough typed ability authority to validate the declared derivation. Prefer an explicit `ability: AbilityName` on `SkillBonus`; alternatively use a closed canonical skill mapping with a clear explicit-override path for nonstandard skills. Do not silently guess unknown strings.
3. Reject duplicate saving-throw abilities and duplicate normalized skill names with stable issue codes and exact list-item paths.
4. Add focused persistence-mode tests for incorrect standard/expertise values, accepted explicit overrides, and duplicates.
5. Add the resulting stable issue codes and any contract change to the PR15 handoff and regenerated schema artifacts.

This belongs in PR14 rather than later hardening because PR15 is instructed to treat a persistence-ready receipt as durable truth, and the active contract explicitly lists deterministic derivation checks as a validation layer.

### Non-blocking metadata cleanup

The PR description is still stale: it advertises 79 tests and says the digest API accepts `CanonicalDefinitionJSON`, while the current public digest API accepts only `StatblockDefinitionV1`. Please refresh the test count and final API summary before merge.

No GitHub status checks are attached to this head.

Drakosfire:
Addressed the proficiency trust-core blocker:

1. Added required `SkillBonus.ability` so skill derivation checks use typed ability authority (no string guessing).
2. Persistence validation now enforces `standard = mod + PB` and `expertise = mod + 2×PB` for saving throws and skills; `explicit_override` is preserved.
3. Rejects duplicate saving-throw abilities and case-normalized duplicate skill names with stable codes and exact list-item paths (`DUPLICATE_SAVING_THROW_ABILITY`, `DUPLICATE_SKILL_NAME`, `SAVING_THROW_DERIVATION_MISMATCH`, `SKILL_DERIVATION_MISMATCH`).
4. Passive Perception uses a unique Perception skill only (duplicates no longer pick the first entry).
5. Regenerated schema artifacts; bumped validator to `statblock-validator-v1.1`; updated PR14/PR15 handoffs and contract docs.

Focused lane: **90 passed** via `./scripts/run_statblocks_v1_tests.sh`. PR description refreshed (digest API accepts only `StatblockDefinitionV1`).

Drakosfire (commented):
## Review verdict: one narrow normalization blocker remains

The latest commit correctly addresses the requested proficiency trust-core work:

- `SkillBonus.ability` is now required typed authority;
- `standard` and `expertise` saving-throw/skill values are checked against ability modifiers and `challenge.proficiency_bonus`;
- `explicit_override` values remain accepted;
- duplicate saving-throw abilities and case-folded skill names are rejected at exact list-item paths;
- validator version, design docs, generated schemas, PR15 handoff, tests, and PR description were updated coherently.

The arithmetic and enum-based ability lookup are sound.

### Required: use canonical Unicode normalization when detecting duplicate skill names

Duplicate skills are currently normalized with only:

```python
normalized = skill.skill.casefold()
```

But the canonicalizer later NFC-normalizes every string before digest and persistence. That means two distinct input strings such as composed and decomposed forms of the same skill name can pass validation as different authorities, then become identical in the canonical persisted definition.

For example, names equivalent to:

```python
"Café"
"Cafe\u0301"
```

are different under the current duplicate check but both canonicalize to `"Café"`. The resulting immutable definition can therefore contain two canonically identical skill authorities despite a persistence-ready receipt.

Required:

1. Normalize skill identity using the same Unicode equivalence as canonicalization before case folding, e.g. NFC + `casefold()` (preferably through a shared normalization helper so the validator and canonicalizer cannot drift).
2. Add a persistence-mode regression test with composed/decomposed equivalent skill names, asserting `DUPLICATE_SKILL_NAME` at the later entry's `.skill` path and `is_persistence_ready == False`.
3. Clarify in the contract/handoff that normalized skill identity means canonical Unicode normalization plus case folding, not case folding alone.

No new issue code or schema change is necessary.

### Verification note

The PR reports 90 isolated tests passing and the metadata is now current. I could not independently execute the suite in this environment because outbound DNS access to GitHub was unavailable. No GitHub status checks or workflow runs are attached to this head.

Drakosfire:
Addressed the NFC duplicate-skill blocker: skill identity now uses Unicode NFC + `casefold()`, matching canonicalization, with a regression test for `Café` vs `Cafe\u0301`. Focused lane: 91 passed.

Drakosfire (commented):
## Review verdict: ready to merge

The latest commit correctly closes the final canonicalization-boundary blocker.

### Verified

- Skill identity now uses NFC normalization followed by case folding for duplicate detection.
- Passive-Perception lookup uses the same normalized identity path, avoiding divergent handling of `Perception`.
- Composed and decomposed Unicode-equivalent skill names are rejected before canonicalization/persistence.
- The issue remains `DUPLICATE_SKILL_NAME` with the exact later-entry path.
- The PR15 handoff and PR description accurately document the final behavior and current 91-test count.

I also rechecked the accumulated trust-core changes at the current head:

- persistence readiness is mode-safe;
- canonical digests accept only parsed `StatblockDefinitionV1` models;
- proficiency, passive Perception, usage, spellcasting, resources, attack ranges/targets, references, action economy, and bounded rules-text contradictions are validated before immutable persistence;
- validation paths are UI-addressable;
- schema artifacts and successor handoff reflect the final contract.

I found no remaining merge blocker in PR14. GitHub reports the PR as mergeable.

No GitHub status checks or workflow runs are attached to this head, so this verdict relies on code/fixture/test inspection and the advertised isolated run of 91 passing tests rather than independently observed CI.

PR #15, opened 2026-07-18
feat(statblocks_v1): PR15 repositories and immutable persistence

Drakosfire:
## Summary
- Rebases PR15 onto merged PR14/`main` and uses the final digest API (`compute_definition_digest(definition)` only).
- Create/append idempotency pins exact revision outcomes; append is CAS; transactional deadline/transport failures reconcile as indeterminate.
- Commands snapshot immutable intent and fix `request_digest` once; in-memory persistence locks all reads/writes; candidate create failures map to `PersistenceUnavailableError` (no `NameError`).
- Firestore keeps native timestamp TTL and has emulator-gated coverage plus unit seams for failure classification.

## Test plan
- [x] `./scripts/run_statblocks_v1_tests.sh` (110 passed, 4 emulator-gated skipped without host)
- [x] Candidate write deadline/general failure → `PersistenceUnavailableError`
- [x] Concurrent readers observe only committed latest/revision/list state
- [x] Mutating source definition/provenance/assets after command construction cannot diverge stored truth or digest

### DIFF
--- a/Docs/Plans/HANDOFF-pr15-statblock-v1-repositories-persistence.md
+++ b/Docs/Plans/HANDOFF-pr15-statblock-v1-repositories-persistence.md
@@ -243,8 +243,10 @@ Define typed domain/application errors for at least:
 - stale parent revision (CAS failure);
 - immutable revision/resource conflict;
 - idempotency conflict;
-- persistence unavailable;
-- transaction indeterminate/reconciliation required.
+- persistence unavailable (non-transaction reads/connectivity);
+- transaction indeterminate/reconciliation required (deadline/transport after a
+  transactional create/append attempt; reconcile via idempotency or fail closed);
+- ambiguous request payload (NFC-colliding map keys in request digests).
 
 Infrastructure exceptions must not leak as raw HTTP details later.
 
--- a/Docs/Plans/HANDOFF-pr15-statblock-v1-repositories-persistence.md
+++ b/Docs/Plans/HANDOFF-pr15-statblock-v1-repositories-persistence.md
@@ -182,9 +182,11 @@ Atomically:
 1. verify idempotency key;
 2. load logical statblock;
 3. verify parent revision exists and belongs to statblock;
-4. write new immutable revision;
-5. update chronological `latest_revision_id`;
-6. write idempotency outcome.
+4. compare-and-swap: submitted parent must equal current `latest_revision_id`
+   (stale parents raise `stale_parent_revision`, distinct from missing/foreign parent);
+5. write new immutable revision;
+6. update chronological `latest_revision_id`;
+7. write idempotency outcome.
 
 Do not overwrite a revision document even when the caller supplies the same ID. IDs are server-owned.
 
@@ -194,7 +196,7 @@ Required behavior:
 
 ```text
 same key + same request digest
-  → return original successful outcome
+  → return original successful outcome (exact statblock_id + revision_id)
 
 same key + different request digest
   → typed idempotency conflict
@@ -203,7 +205,13 @@ retry after partial infrastructure error
   → reconcile transaction outcome before creating anything new
 ```
 
-The request digest is not the definition digest alone; it should distinguish create/append operation parameters and parent/candidate intent.
+Create outcomes must pin the original created revision, not `latest_revision_id`.
+Replay checks the idempotency record before persistence validation/ID allocation, with
+an atomic recheck at commit.
+
+The request digest is not the definition digest alone. It hashes operation parameters
+with the definition component as PR14 canonical JSON and NFC-normalized remaining
+strings so Unicode/set-order equivalents do not false-conflict.
 
 ## 6. Candidate persistence
 
@@ -232,7 +240,8 @@ Define typed domain/application errors for at least:
 - statblock not found;
 - revision not found;
 - parent revision mismatch;
-- immutable revision conflict;
+- stale parent revision (CAS failure);
+- immutable revision/resource conflict;
 - idempotency conflict;
 - persistence unavailable;
 - transaction indeterminate/reconciliation required.
--- a/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
+++ b/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
@@ -15,10 +15,15 @@
 - Candidate IDs use `cand_<base36>`; statblock/revision IDs use `sb_<base36>` and
   `rev_<base36>`. Candidate TTL is enforced by `expires_at`; Firestore TTL must also
   be configured on that field.
-- `compute_request_digest(operation, payload)` covers complete operation parameters;
-  it is distinct from the definition digest. Firestore collections are
-  `dungeonbuddy_statblock_candidates_v1`, `dungeonbuddy_statblocks_v1` with
-  `revisions` subcollections, and `dungeonbuddy_statblock_idempotency_v1`.
+- `compute_request_digest(operation, payload)` covers complete operation parameters
+  with the definition component as PR14 canonical JSON plus NFC-normalized remaining
+  strings; it is distinct from the definition digest.
+  Create/append idempotency outcomes pin `IdempotencyOutcomeV1(statblock_id, revision_id)`
+  so create replay returns the original revision after later appends.
+  Append is compare-and-swap against `latest_revision_id` (`stale_parent_revision`).
+  Firestore collections are `dungeonbuddy_statblock_candidates_v1`,
+  `dungeonbuddy_statblocks_v1` with `revisions` subcollections, and
+  `dungeonbuddy_statblock_idempotency_v1`. Candidate `expires_at` is a native timestamp.
 
 ## 0. Mission
 
--- a/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
+++ b/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
@@ -5,6 +5,21 @@
 **Predecessors:** PR13 contract, PR14 trust core, PR15 repository protocols  
 **Successor:** `HANDOFF-pr17-statblock-v1-candidate-api.md`
 
+## PR15 predecessor completion notes
+
+- Use synchronous `CandidateRepository.create/get` and `StatblockPersistenceRepository`
+  (`create_statblock`, `append_revision`, `get`, `get_revision`); async callers must
+  offload calls with `asyncio.to_thread`.
+- Offline fixtures can use `InMemoryCandidateRepository` and
+  `InMemoryStatblockPersistenceRepository(clock=..., id_factory=DeterministicIdFactory())`.
+- Candidate IDs use `cand_<base36>`; statblock/revision IDs use `sb_<base36>` and
+  `rev_<base36>`. Candidate TTL is enforced by `expires_at`; Firestore TTL must also
+  be configured on that field.
+- `compute_request_digest(operation, payload)` covers complete operation parameters;
+  it is distinct from the definition digest. Firestore collections are
+  `dungeonbuddy_statblock_candidates_v1`, `dungeonbuddy_statblocks_v1` with
+  `revisions` subcollections, and `dungeonbuddy_statblock_idempotency_v1`.
+
 ## 0. Mission
 
 Implement the provider-independent application service that generates and revises `StatblockDefinitionV1` using OpenAI Structured Outputs, validates the result, persists a server-owned candidate record, and returns `GeneratedStatblockCandidateV1`.
--- a/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
+++ b/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
@@ -5,6 +5,19 @@
 **Predecessors:** repository persistence and generation service  
 **Successor:** `HANDOFF-pr18-statblock-v1-revision-resource-api.md`
 
+## PR15 predecessor completion notes
+
+- Use the synchronous repository protocols from `statblocks_v1.application.repositories`;
+  route/service code must use `await asyncio.to_thread(...)` for Firestore calls.
+- `InMemoryCandidateRepository` and `InMemoryStatblockPersistenceRepository` accept
+  deterministic clock and ID fixtures. IDs are `cand_<base36>`, `sb_<base36>`, and
+  `rev_<base36>`.
+- Candidates expire at `expires_at` (also the Firestore TTL field). Collections are
+  `dungeonbuddy_statblock_candidates_v1`, `dungeonbuddy_statblocks_v1/revisions`,
+  and `dungeonbuddy_statblock_idempotency_v1`.
+- Reuse `compute_request_digest(operation, payload)` for request replay: it includes
+  operation parameters rather than using only the mechanics definition digest.
+
 ## 0. Mission
 
 Expose the generation, revision, validation, and candidate-read operations through the dedicated authenticated DungeonBuddy v1 router.
--- a/pytest.ini
+++ b/pytest.ini
@@ -13,3 +13,4 @@ markers =
     live: marks tests that require live API calls
     unit: marks tests as unit tests
     asyncio: marks tests as async tests
+    firestore_emulator: requires FIRESTORE_EMULATOR_HOST and a Firestore emulator
--- a/statblocks_v1/application/repositories.py
+++ b/statblocks_v1/application/repositories.py
@@ -9,6 +9,7 @@ import hashlib
 import json
 import unicodedata
 from collections.abc import Callable, Mapping
+from copy import deepcopy
 from datetime import datetime
 from typing import Any, Protocol
 
@@ -95,6 +96,19 @@ class StatblockPersistenceRepository(
 
 
 class CreateStatblockCommand:
+    """Immutable create intent: inputs are snapshotted and the digest is fixed."""
+
+    __slots__ = (
+        "caller_scope",
+        "idempotency_key",
+        "created_by",
+        "candidate_id",
+        "_definition",
+        "_provenance",
+        "_asset_bindings",
+        "_request_digest",
+    )
+
     def __init__(
         self,
         *,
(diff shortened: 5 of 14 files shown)


Drakosfire (commented):
## Review verdict

Reviewed head `04e80445301cfc30530b26d1ccc57e7b150dca52` against the merged PR14 contract (`f9a5cc741204ada50d3ab35b6107ef2869688081`). PR15 is still based on the early PR14 branch at `4083ca196dad964a589329b6e440486d32455a62`, is currently non-mergeable, and no status checks or workflow runs are attached to the head.

The PR gets several important boundaries right: repository methods are explicitly synchronous, v1 collections are isolated from legacy storage, statblock/revision IDs are server-allocated, accepted definitions are persistence-validated, revision documents use create semantics in Firestore, and no HTTP routes are introduced. The remaining defects are at the durable-truth boundary, so successor PRs cannot safely trust this implementation yet.

## Required changes

### 1. Rebase onto merged PR14 and consume the final trust-core API

The branch does not contain the final reviewed PR14 contract. In particular, `memory_repositories._persistence_material()` calls `compute_definition_digest(canonical)`, relying on the earlier API that accepted canonical strings. Merged PR14 intentionally accepts **only** `StatblockDefinitionV1`; after a correct rebase this call raises `TypeError`.

Required correction:

- rebase/retarget PR15 onto current `main` containing PR14 merge `f9a5cc7`;
- compute `canonical = canonicalize_definition(definition)` and `digest = compute_definition_digest(definition)` through the final APIs;
- persist that exact canonical text, digest, and persistence receipt together;
- retain every final PR14 schema/fixture/validation correction rather than resolving conflicts toward this branch’s older copies.

Focused proof:

- create and append succeed on the rebased implementation;
- `compute_definition_digest` still rejects strings/bytes/canonical wrappers;
- stored canonical text, revision digest, and `receipt.definition_digest` agree exactly.

### 2. Preserve the exact original idempotency outcome and canonicalize the request digest

Create replay currently stores only a statblock locator. Both implementations then reconstruct the returned revision from `statblock.latest_revision_id`. This permits:

```text
create with key K -> returns revision R1
append revision R2
retry the original create with key K -> returns R2
```

That is not replay of the original successful outcome.

The Firestore implementation also validates, canonicalizes, allocates IDs, and builds resources before checking the idempotency record inside the transaction. A successful historical request can therefore fail on replay after validator behavior changes instead of returning the durable outcome.

Finally, `compute_request_digest()` hashes raw `model_dump()` JSON. It does not use PR14 canonicalization or NFC normalization, so canonically equivalent definitions—including normalized Unicode and set-like field ordering—can conflict under the same idempotency key.

Required correction:

- persist an operation-specific outcome that pins the exact logical statblock **and exact created revision**;
- replay that exact outcome even after later appends;
- check/reconcile idempotency before persistence validation and allocation, with an atomic recheck at commit;
- canonicalize the definition component and define deterministic Unicode/omitted/null/order semantics for the whole request digest.

Focused proof:

- create → append → retry create returns R1, not latest;
- replay does not invoke a changed/failing validator or allocate new IDs;
- NFC-equivalent and canonically equivalent definitions produce the same request digest;
- changed parent, candidate provenance, assets, or other intent still produces an idempotency conflict.

### 3. Make append a compare-and-swap against the current parent

Both repositories verify only that the named parent exists. They do not verify that it is still the logical statblock’s current `latest_revision_id`. Two appends from the same parent can therefore both succeed; Firestore retries the second transaction after contention and then blindly advances `latest_revision_id` to a sibling revision.

Required correction:

- inside the in-memory lock / Firestore transaction, require the submitted parent to equal the current latest revision before creating the child and advancing the pointer;
- return a typed stale-parent/concurrency conflict distinct from missing statblock and cross-statblock parent mismatch.

Focused proof:

- two different idempotency keys append from the same parent: exactly one succeeds;
- the rejected append creates no revision or idempotency outcome;
- missing statblock, missing parent, foreign parent, and stale parent remain distinguishable;
- emulator coverage exercises real concurrent append behavior.

### 4. Prevent mutation and overwrite of in-memory historical truth

The stored in-memory revision is constructed with `definition=command.definition`. Pydantic retains an already-validated nested model instance, so a caller can mutate the original definition after insertion and mutate the repository’s stored `revision.definition` while `canonical_definition`, `definition_digest`, and the persisted receipt remain unchanged. That creates multiple contradictory co-authoritative representations inside one revision.

There are also semantic parity gaps: `InMemoryCandidateRepository.create()` overwrites an existing candidate ID, and revision assignment can overwrite an existing revision if an injected ID factory collides, while Firestore uses create-only writes.

Required correction:

- deep-copy or serialize/revalidate every mutable input at the write boundary before storing it;
- reject candidate/revision ID collisions with typed immutable-conflict behavior; never upsert immutable records;
- preserve the existing deep-copy behavior on reads.

Focused proof:

- mutating the submitted definition, provenance, asset bindings, or candidate after create does not change stored state;
- mutating a returned object does not change stored state;
- canonicalizing the stored parsed definition still equals stored canonical text and digest;
- duplicate candidate IDs and forced revision-ID collisions cannot overwrite history.

### 5. Implement and prove the Firestore contract rather than only its collection constants

`_dump(model_dump(mode="json"))` converts `datetime` fields to strings. Firestore TTL requires `expires_at` to be stored as a timestamp, so the documented TTL policy cannot act on these candidate documents.

The statblock/revision/list/idempotency read methods also allow raw Firestore exceptions to cross the repository boundary, and append maps a missing logical statblock to `ParentRevisionMismatchError`. Those behaviors disagree with both the declared failure model and the in-memory substitute.

The emulator-gated test currently asserts only three collection-name constants. It does not exercise any Firestore write, transaction, round trip, TTL field, idempotency replay, contention, or failure translation required by the handoff.

Required correction:

- use a Firestore serializer that preserves timestamp values while retaining exact canonical definition text;
- translate infrastructure failures consistently; reserve `TransactionIndeterminateError` for genuinely unknown commit outcomes and distinguish known unavailable/pre-commit failures;
- return `StatblockNotFoundError` when the logical statblock is absent;
- add actual emulator/failure-injection coverage for atomic create, exact round trip, timestamp TTL field, append contention, create/append replay, immutable create preconditions, and reconciliation after an unknown commit outcome.

Ordinary CI can continue skipping emulator tests without credentials, but the complete in-memory contract suite must prove the same significant semantics.

## Non-blocking observation

`list_for_statblock()` currently provides no explicit ordering contract. That is acceptable if successors treat it as an unordered collection; otherwise define chronological ordering now and prove identical behavior in memory and Firestore.

Drakosfire:
Addressed the blocking durable-truth review:

1. Rebased onto merged PR14/`main` and retargeted the PR base to `main`. Digests use `compute_definition_digest(definition)` only; stored canonical text, revision digest, and receipt digest agree.
2. Idempotency outcomes are now `IdempotencyOutcomeV1(statblock_id, revision_id)`. Create→append→retry create returns the original revision. Replay checks the record before validation/allocation; request digests use PR14 canonical definition JSON + NFC normalization.
3. Append is CAS against `latest_revision_id` with typed `stale_parent_revision`, distinct from missing statblock / missing-or-foreign parent.
4. In-memory writes deep-copy mutable inputs; candidate/statblock/revision ID collisions raise immutable-conflict errors instead of overwriting.
5. Firestore `_dump` preserves native timestamps for TTL; reads translate infra failures; missing logical statblock returns `StatblockNotFoundError`. Emulator-gated tests cover create replay, TTL timestamp round-trip, concurrent append CAS, and missing-statblock distinction. In-memory suite proves the same significant semantics in ordinary CI.

Focused lane: **102 passed**, 4 emulator tests skipped without `FIRESTORE_EMULATOR_HOST`.

Drakosfire (commented):
## Re-review verdict

**BLOCKING — substantially improved, but not ready to merge.**

Reviewed head `f44aa883cc2edb01fd493c22fd3bd419d4e47f26`, now correctly based on merged PR14 / `main` at `f9a5cc741204ada50d3ab35b6107ef2869688081`. The PR is mergeable. No GitHub status checks or workflow runs are attached to the head. The PR body reports `102 passed, 4 skipped`; I could not independently execute the suite because this review environment could not resolve GitHub for a checkout.

## Prior finding ledger

1. **Rebase/final PR14 API — closed.** The branch is based on merged PR14, `_persistence_material()` now calls `compute_definition_digest(definition)`, and tests preserve the definition-only digest boundary.
2. **Exact idempotency replay — closed for the ordinary path.** `IdempotencyOutcomeV1` pins both IDs; create → append → retry create returns the original revision. Definition request intent is canonicalized and remaining strings are NFC-normalized.
3. **Stale-parent append — closed.** Both adapters compare the submitted parent to `latest_revision_id`, and the emulator suite exercises concurrent sibling attempts.
4. **Mutation/overwrite isolation — closed for statblock/revision writes.** Inputs and reads are deep-copied, candidate/revision collisions are rejected, and focused tests verify historical state remains coherent.
5. **Firestore timestamps and basic emulator proof — mostly closed.** `_dump()` preserves native datetimes; the emulator suite now covers candidate TTL round-trip, exact revision round-trip, replay, concurrent append, and missing-statblock behavior.

The remaining blockers are narrower, but they still sit at the repository trust boundary.

## Required changes

### 1. Transaction transport failures still erase the distinction between unavailable and indeterminate outcomes

`_run_transaction()` classifies `ServiceUnavailable`, `DeadlineExceeded`, `Unavailable`, and `RetryError` as `PersistenceUnavailableError`. For a transactional create/append, those exceptions may be observed after Firestore committed the transaction but before the client received a definitive response. That is exactly the `transaction_indeterminate` / reconciliation-required state the handoff requires.

The current behavior permits this sequence:

```text
transaction commits statblock + revision + idempotency outcome
client receives DeadlineExceeded
repository raises persistence_unavailable
```

The durable result exists, but the typed failure tells the successor service only that storage is unavailable, not that it must reconcile before deciding what happened. The new tests do not inject or prove this commit-then-error path.

Required correction:

- treat transaction-attempt transport/deadline/retry-exhaustion failures as potentially indeterminate unless the adapter can prove they happened before commit;
- reconcile through the scoped idempotency record and return the original outcome when it is observable;
- if reconciliation cannot establish the outcome, raise `TransactionIndeterminateError`, not `PersistenceUnavailableError`;
- keep ordinary non-transaction read/connectivity failures mapped to `PersistenceUnavailableError`.

Focused proof:

- simulate a transaction that durably writes its outcome and then raises a deadline/transport error;
- the repository reconciles and returns the exact original statblock/revision;
- when the reconciliation read is also unavailable, it raises `TransactionIndeterminateError`;
- retrying the same command allocates no additional IDs and creates no duplicate resources.

### 2. `InMemoryCandidateRepository.create()` is not atomic despite claiming Firestore parity

The candidate repository performs `if candidate_id in self._candidates` followed by assignment without a lock. Two concurrent creates of the same server-owned candidate ID can both observe absence, both return success, and one silently replaces the other. Firestore uses create-only semantics and allows only one success.

This leaves the deterministic substitute materially weaker than the production adapter and means successor concurrency tests can pass in memory while the real store behaves differently.

Required correction:

- protect candidate create/get state with a lock, or otherwise make the check-and-create atomic;
- preserve deep-copy isolation inside that atomic boundary.

Focused proof:

- two concurrent creates for the same candidate ID produce exactly one success and one `ImmutableResourceConflictError`;
- the stored candidate equals the winning immutable record and is not affected by either caller afterward.

### 3. NFC normalization can collapse distinct request-map keys into the same idempotency digest

`_normalize_request_payload()` normalizes mapping keys inside a dict comprehension. Distinct Python keys such as `"é"` and `"e\u0301"` normalize to the same JSON key, so one silently overwrites the other during digest construction.

For example, these materially different provenance payloads can produce the same request digest:

```python
{"é": 1}
{"é": 1, "e\u0301": 2}
```

The persisted provenance is not normalized the same way, so changed operation intent can bypass the required idempotency conflict and replay an unrelated prior outcome.

Required correction:

- detect collisions after key normalization and fail closed rather than overwriting a value;
- define the stable typed error or command-validation behavior for an ambiguous request map.

Focused proof:

- composed/decomposed **values** remain equivalent;
- composed/decomposed duplicate **keys** are rejected;
- adding or changing any non-equivalent provenance/asset intent changes the request digest.

### 4. Firestore immutable-collision errors lose the owning resource and can report the wrong type

Any `AlreadyExists` raised from `_run_transaction()` becomes `ImmutableRevisionConflictError("unknown")`. During first-create this may be the logical statblock document, not a revision; during append the actual colliding revision ID is known but discarded.

Required correction:

- map a first-create statblock collision to `ImmutableResourceConflictError("statblock", statblock_id)`;
- map an append revision collision to `ImmutableRevisionConflictError(revision_id)`;
- retain the real identifier in structured error details.

Focused proof:

- forced statblock-ID collision and forced revision-ID collision produce distinct typed errors with the exact IDs;
- neither collision writes an idempotency outcome or advances `latest_revision_id`.

## Non-blocking observations

- The emulator tests use fixed candidate IDs and idempotency keys without cleanup, so repeated runs against the same emulator instance can fail or replay state from an earlier run. Unique per-test namespaces or teardown would make the suite reliably rerunnable.
- `list_for_statblock()` remains intentionally unordered; successors should not infer chronological ordering from either implementation.

Drakosfire:
Addressed the four remaining blockers:

1. **Indeterminate transactions** — deadline/transport/retry failures from transactional create/append now raise `TransactionIndeterminateError` (not `persistence_unavailable`). Create/append reconcile through the scoped idempotency record and return the exact original outcome when observable; if the reconcile read is also unavailable or missing, they stay indeterminate.
2. **Candidate create race** — `InMemoryCandidateRepository` now locks check-and-create; concurrent same-ID creates yield exactly one success and one `ImmutableResourceConflictError`.
3. **NFC key collapse** — request-digest normalization fails closed with `AmbiguousRequestPayloadError` when distinct keys collide after NFC; NFC-equivalent values still digest equally.
4. **AlreadyExists typing** — create collisions map to `ImmutableResourceConflictError("statblock", id)`; append collisions map to `ImmutableRevisionConflictError(revision_id)`.

Focused lane: **108 passed**, 4 emulator tests skipped without `FIRESTORE_EMULATOR_HOST`.

Drakosfire (commented):
## Re-review verdict

**BLOCKING — close, but not ready to merge.**

Reviewed head `af0ea7e6bdf1ad32ed4b5b5cc982613f868b3d75` against the prior reviewed head `f44aa883cc2edb01fd493c22fd3bd419d4e47f26` and then rechecked the complete PR15 persistence invariant.

The new commit correctly closes the four findings from review `4729805539`:

- transactional non-domain failures are now treated as indeterminate and reconciled through the durable idempotency outcome;
- in-memory candidate create/get now share an `RLock`;
- NFC-colliding request-map keys fail closed with a typed error;
- Firestore `AlreadyExists` mapping now preserves statblock/revision resource identity.

Three durable-boundary issues remain.

### 1. Firestore candidate write failures now raise `NameError`

`FirestoreCandidateRepository.create()` still calls `_is_unavailable(error)`, but this commit deletes `_is_unavailable()` from the module. Any candidate write failure that is not classified as `AlreadyExists` now masks the original Firestore error with:

```text
NameError: name '_is_unavailable' is not defined
```

That violates the typed failure boundary and means the new test count does not cover this path.

Required correction:

- either restore the helper or remove the dead branch and map all non-duplicate candidate write failures directly to `PersistenceUnavailableError`;
- add a failure-injection unit test proving an unavailable/deadline/general client exception cannot leak or become `NameError`.

### 2. The in-memory persistence repository is not thread-safe or transactionally observable

The module and class promise a thread-safe substitute whose single lock models Firestore atomicity, but only `create_statblock()` and `append_revision()` acquire `_lock`. These methods remain unlocked:

- `get()`;
- `get_idempotency()`;
- `get_revision()`;
- `list_for_statblock()`.

Writes publish statblock, revision, latest pointer, and idempotency record in several Python assignments while holding the lock. Because readers do not acquire that lock, another thread can observe an impossible partial transaction—for example a statblock whose `latest_revision_id` is not yet readable, a new revision before latest/idempotency advances, or latest before the idempotency record exists. `list_for_statblock()` can also iterate `_revisions` while another thread mutates it and raise `RuntimeError: dictionary changed size during iteration`.

Required correction:

- guard every in-memory persistence read with the same `RLock`;
- add a synchronized concurrent read/write test proving readers observe only pre-commit or post-commit state and list iteration cannot race.

Using `RLock` preserves the existing replay path, which calls read methods while already inside the write lock.

### 3. One logical operation recomputes intent from mutable command objects

`CreateStatblockCommand` and `AppendRevisionCommand` retain mutable references to the definition, provenance, and asset bindings, and `request_digest` is a computed property. Both repositories read these fields repeatedly across replay checks, validation/canonicalization, deep-copy storage, transaction execution, reconciliation, and idempotency recording.

A caller mutation concurrent with the synchronous/offloaded repository call can therefore produce contradictory durable truth:

- canonical text/digest/receipt computed from definition A while the stored parsed definition is copied from definition B;
- persisted provenance/assets from one state while the idempotency digest is recomputed from another;
- different digests used for the precheck, transaction record, and reconciliation.

This is especially relevant because the documented API passes these synchronous calls across `asyncio.to_thread`.

Required correction:

- create one defensive, immutable operation snapshot at the boundary;
- compute the request digest exactly once from that snapshot;
- use the same snapshot and digest for validation, resource construction, transaction, reconciliation, and idempotency recording in both adapters;
- add a barrier-controlled mutation test proving the stored parsed definition, canonical text, definition digest, receipt, provenance/assets, and request digest cannot diverge.

## Verification note

GitHub reports the PR mergeable and correctly based on merged PR14. No status checks or workflow runs are attached to this head, so the claimed `108 passed` result is not independently visible here.

Because this is a self-owned PR, this is submitted as a blocking `COMMENT` review rather than `REQUEST_CHANGES`.

Drakosfire:
Addressed the three remaining blockers:

1. **Candidate `NameError`** — removed the dead `_is_unavailable` branch; non-duplicate candidate write failures map directly to `PersistenceUnavailableError`, with a failure-injection unit test.
2. **In-memory read/write atomicity** — `get`, `get_idempotency`, `get_revision`, and `list_for_statblock` now share the persistence `RLock`; concurrent reader/writer coverage proves committed-state observability without dict-iteration races.
3. **Immutable operation intent** — `CreateStatblockCommand` / `AppendRevisionCommand` deep-copy inputs at construction, freeze `request_digest` once, and reject attribute reassignment; repositories consume that single snapshot/digest end-to-end.

Focused lane: **110 passed**, 4 emulator tests skipped without `FIRESTORE_EMULATOR_HOST`.

Drakosfire (commented):
## Re-review verdict

Reviewed head `54ed5e6e31de8f9b13bff1b77304d09fef7c8adb` against the prior reviewed head `af0ea7e6bdf1ad32ed4b5b5cc982613f868b3d75`, then rechecked the complete PR15 persistence invariant rather than only the new diff.

The latest commit closes all three findings from review `4729889645`:

1. **Candidate failure regression — closed.** `FirestoreCandidateRepository.create()` now maps every non-duplicate client failure directly to `PersistenceUnavailableError`; the deleted `_is_unavailable()` helper is no longer referenced, and a focused failure-injection test covers the path.
2. **In-memory transaction visibility — closed.** `get`, `get_idempotency`, `get_revision`, and `list_for_statblock` now acquire the same `RLock` as create/append. Readers can no longer observe the multi-dictionary write sequence mid-commit or race revision iteration.
3. **Mutable operation intent — closed.** Create/append commands deep-snapshot the definition, provenance, and asset bindings at construction, fix the request digest once, reject public attribute mutation, and return defensive copies. Both adapters snapshot those command properties once and reuse the same digest through replay checks, validation, resource construction, transaction recording, and reconciliation.

## Full persistence ledger

The complete PR now satisfies the reviewed PR15 contract:

- based on merged PR14 and uses the final definition-only digest API;
- exact idempotency outcomes pin both `statblock_id` and the original `revision_id`;
- create replay after later appends returns the original first revision;
- append is compare-and-swap against the current latest revision;
- stale, missing, and foreign parents remain typed and distinct;
- immutable candidate/statblock/revision collisions cannot overwrite history;
- in-memory inputs, stored state, and returned values are mutation-isolated;
- request digests use canonical definition JSON, NFC-normalized values, and reject normalized-key collisions;
- Firestore candidate TTL fields remain native timestamps;
- transactional transport uncertainty reconciles through durable idempotency and otherwise fails closed as `transaction_indeterminate`;
- Firestore `AlreadyExists` mapping preserves the actual resource type and ID;
- real emulator-gated tests cover round-trip persistence, TTL representation, replay, and concurrent append behavior;
- repository methods remain synchronously explicit, with `asyncio.to_thread` required at async boundaries;
- no final HTTP resource route is introduced.

## Verification note

GitHub reports this PR mergeable and correctly based on `main`. No status checks or workflow runs are attached to this head, so the PR body's reported `110 passed, 4 emulator-gated skipped` result is not independently visible in GitHub from this review environment.

## Non-blocking follow-up

The emulator tests still use fixed IDs/keys and may not be safely rerunnable against a reused emulator instance without teardown or unique namespaces. `list_for_statblock()` also remains intentionally unordered; successors should not infer chronology from its return order.

Because this is a self-owned PR, this ready verdict is submitted as a `COMMENT` review rather than `APPROVE`.

PR #16, opened 2026-07-18
feat(statblocks_v1): PR16 structured outputs generation service

Drakosfire:
## Summary
- Provider-independent `GenerationServiceV1` generates/revises `StatblockDefinitionV1` through a typed provider seam, validates in candidate mode, and persists server-owned candidates (no HTTP routes).
- Trust boundaries: ruleset match, verified source digests, server-owned caller provenance, exact `(statblock_id, revision_id)` locators, key-preservation warnings, typed provider/resolver failures, in-repo `MODEL_POLICY.json`.
- Offline fake-provider matrix plus opt-in live OpenAI Structured Outputs smoke tests.

## Test plan
- [x] `./scripts/run_statblocks_v1_tests.sh` (isolated lane)
- [x] Revision from definition and exact locator
- [x] Warning-bearing candidate persistence
- [x] Provider outcome catalog + unexpected provider exceptions
- [x] Ruleset mismatch / source digest mismatch
- [ ] Optional: `RUN_OPENAI_GENERATION_TESTS=1 OPENAI_API_KEY=... ./scripts/run_statblocks_v1_tests.sh -k openai`

### DIFF
--- a/Docs/Design/DESIGN-dungeonbuddy-statblock-contract-v1.md
+++ b/Docs/Design/DESIGN-dungeonbuddy-statblock-contract-v1.md
@@ -206,6 +206,49 @@ GeneratedStatblockCandidateV1
   generation_receipt
   asset_brief
   assets
+  asset_warnings[]
+  created_at
+  expires_at
+  source_locator
+```
+
+`generation_receipt` is server-owned audit data:
+
+```text
+GenerationReceiptV1
+  request_id
+  provider
+  model
+  prompt_version
+  schema_version
+  schema_fingerprint
+  generated_at
+  caller_scope
+  actor
+  source_description_digest
+  source_definition_digest
+  source_locator
+  provider_request_id / provider_response_id
+  latency_ms / input_tokens / output_tokens
+```
+
+Ownership and provenance rules:
+
+- `caller_scope` / `actor` come from the authenticated request, never from model output.
+- `source_description_digest` digests authored prose (NFC + SHA-256). Callers may supply a
+  matching digest; a mismatch is rejected. It does not identify source mechanics.
+- `source_definition_digest` is required for every revision and is the PR14 digest of the
+  exact `StatblockDefinitionV1` embedded in the revision prompt (inline submission or
+  resolved immutable revision).
+- `source_locator` is present when revision used an exact `(statblock_id, revision_id)` read.
+- When `generate_images` is requested, `asset_brief` stores the exact effective brief passed
+  to the asset generator (authored description brief, or creature-name fallback).
+- `asset_warnings` are typed partial outcomes with stable codes, not free-form strings:
+
+```text
+AssetWarningV1
+  code: asset_generator_unconfigured | asset_generation_failed
+  message
 ```
 
 A candidate is not accepted mechanics truth. It may be edited, regenerated, rejected, or
@@ -915,9 +958,9 @@ The response is another candidate, not a persisted revision.
 
 ```json
 {
-  "contract": "dungeonmind.dungeonbuddy-statblocks",
-  "contract_version": "1.0.0",
-  "candidate_id": "candidate_...",
+  "contract": "dungeonbuddy-statblock",
+  "contract_version": "v1",
+  "candidate_id": "cand_...",
   "definition": {},
   "validation_receipt": {
     "status": "warnings",
@@ -925,16 +968,38 @@ The response is another candidate, not a persisted revision.
   },
   "generation_receipt": {
     "request_id": "req_...",
-    "generator_version": "...",
+    "provider": "openai",
     "model": "...",
-    "prompt_version": "...",
-    "created_at": "..."
+    "prompt_version": "statblock-generation-prompt-v1",
+    "schema_version": "...",
+    "schema_fingerprint": "...",
+    "generated_at": "...",
+    "caller_scope": "dungeonbuddy",
+    "actor": "optional-actor",
+    "source_description_digest": "sha256:...",
+    "source_definition_digest": "sha256:...",
+    "source_locator": {
+      "statblock_id": "sb_...",
+      "revision_id": "rev_..."
+    }
   },
   "asset_brief": {
     "prompt": "...",
     "recommended_roles": ["portrait", "token"]
   },
-  "assets": []
+  "assets": [],
+  "asset_warnings": [
+    {
+      "code": "asset_generator_unconfigured",
+      "message": "Asset generation was requested but no asset generator is configured."
+    }
+  ],
+  "created_at": "...",
+  "expires_at": "...",
+  "source_locator": {
+    "statblock_id": "sb_...",
+    "revision_id": "rev_..."
+  }
 }
 ```
 
--- a/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
+++ b/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
@@ -1,6 +1,6 @@
 # HANDOFF — PR16 Statblock v1 Structured Outputs generation service
 
-**Status:** IN REVIEW — four remaining trust-boundary blockers closed (digest provenance, key repurposing, fail-closed settings, typed assets) 
+**Status:** IN REVIEW — successor-facing gaps closed (pinned operation intent, design sync, Firestore typed round-trip) 
 **Target repository:** `Drakosfire/DungeonMindServer`  
 **Predecessors:** PR13 contract, PR14 trust core, PR15 repository protocols  
 **Successor:** `HANDOFF-pr17-statblock-v1-candidate-api.md`
--- a/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
+++ b/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
@@ -1,6 +1,6 @@
 # HANDOFF — PR16 Statblock v1 Structured Outputs generation service
 
-**Status:** IN REVIEW — rebased onto merged PR15; trust boundaries sealed  
+**Status:** IN REVIEW — four remaining trust-boundary blockers closed (digest provenance, key repurposing, fail-closed settings, typed assets) 
 **Target repository:** `Drakosfire/DungeonMindServer`  
 **Predecessors:** PR13 contract, PR14 trust core, PR15 repository protocols  
 **Successor:** `HANDOFF-pr17-statblock-v1-candidate-api.md`
--- a/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
+++ b/Docs/Plans/HANDOFF-pr16-statblock-v1-structured-generation-service.md
@@ -1,6 +1,6 @@
 # HANDOFF — PR16 Statblock v1 Structured Outputs generation service
 
-**Status:** READY AFTER PR14; INTEGRATES PR15  
+**Status:** IN REVIEW — rebased onto merged PR15; trust boundaries sealed  
 **Target repository:** `Drakosfire/DungeonMindServer`  
 **Predecessors:** PR13 contract, PR14 trust core, PR15 repository protocols  
 **Successor:** `HANDOFF-pr17-statblock-v1-candidate-api.md`
--- a/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
+++ b/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
@@ -37,17 +37,28 @@
   `definition_invalid`, `ruleset_mismatch`, `source_digest_mismatch`, and
   resolver codes (`revision_not_found`, `statblock_not_found`, `source_unavailable`)
   reserved for service-level failures.
-- Generation receipts store server-owned `caller_scope`/`actor` provenance and a
-  server-computed (optionally caller-verified) source description digest.
+- Generation receipts store server-owned `caller_scope`/`actor` provenance, a
+  server-computed (optionally caller-verified) source description digest, and for
+  every revision a PR14 `source_definition_digest` of the exact mechanics revised
+  (inline submitted definition or resolved locator revision). Description digests
+  never substitute for mechanics provenance.
   Warning-bearing candidate receipts may be persisted; invalid receipts are not.
-  `preserve_element_keys` is validated after revision (`ELEMENT_KEY_CHANGED` /
-  `ELEMENT_KEY_DROPPED` warnings). Requested image generation without a configured
-  asset generator yields an explicit asset warning.
+  `preserve_element_keys` is a versioned post-validation pass
+  (`validator_version` gains `+statblock-key-preservation-v1`) emitting
+  `ELEMENT_KEY_CHANGED`, `ELEMENT_KEY_DROPPED`, `ELEMENT_KEY_REPURPOSED`, and
+  `ELEMENT_KEY_IDENTITY_AMBIGUOUS`.
+  Asset partial outcomes use typed `AssetWarningV1` codes
+  (`asset_generator_unconfigured`, `asset_generation_failed`). When
+  `generate_images=True`, the exact effective `AssetBriefV1` passed to the
+  generator is always persisted on the candidate (name fallback when
+  `include_generation_brief=False`).
 - Settings are `STATBLOCKS_V1_OPENAI_MODEL`,
   `STATBLOCKS_V1_OPENAI_TIMEOUT_SECONDS`, `STATBLOCKS_V1_OPENAI_MAX_RETRIES`, and
-  `STATBLOCKS_V1_CANDIDATE_TTL_SECONDS`. The default model resolves the in-repo
-  `DungeonMindServer/MODEL_POLICY.json` action `structured_generation` (not a path
-  above the repository root); missing policy raises
+  `STATBLOCKS_V1_CANDIDATE_TTL_SECONDS`. Values are validated on the settings
+  model itself (non-empty model, finite timeout `> 0`, retries `>= 0`, TTL `> 0`);
+  malformed env input raises `InternalServiceMisconfiguredError`. The default
+  model resolves the in-repo `DungeonMindServer/MODEL_POLICY.json` action
+  `structured_generation`; missing policy raises
   `InternalServiceMisconfiguredError` unless the env override is set.
 - `FakeDefinitionProvider` accepts a JSON payload or `ProviderOutcomeV1` and records
   calls. It supports offline route tests; `candidate_id_factory` and clock make
--- a/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
+++ b/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
@@ -25,6 +25,10 @@
   `GeneratedStatblockCandidateV1` or `GenerationFailureV1`; routes must map the latter
   without exposing provider exceptions (provider exceptions are caught at the service
   boundary as `provider_failure`).
+- The service deep-copies and pins operation intent before the provider call (prompt,
+  digests, ruleset, caller, locator, source definition, asset options). Concurrent
+  mutation of the caller's command after entry cannot change receipt provenance or
+  key-preservation inputs. Treat commands as write-once at the route boundary.
 - Construct the service through dependency injection with a `DefinitionProvider`,
   `CandidateRepository`, `GenerationSettingsV1`, clock, candidate-ID factory, and,
   for revision locators, `PersistenceDefinitionResolver` over
--- a/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
+++ b/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
@@ -23,18 +23,32 @@
 - Application entry points are `GenerationServiceV1.generate(GenerateStatblockCommandV1)`
   and `GenerationServiceV1.revise(ReviseStatblockCommandV1)`. Both return either
   `GeneratedStatblockCandidateV1` or `GenerationFailureV1`; routes must map the latter
-  without exposing provider exceptions.
+  without exposing provider exceptions (provider exceptions are caught at the service
+  boundary as `provider_failure`).
 - Construct the service through dependency injection with a `DefinitionProvider`,
   `CandidateRepository`, `GenerationSettingsV1`, clock, candidate-ID factory, and,
-  for revision locators, a definition resolver. Tests use `FakeDefinitionProvider`
-  plus `InMemoryCandidateRepository`.
+  for revision locators, `PersistenceDefinitionResolver` over
+  `StatblockPersistenceRepository.get_revision(statblock_id, revision_id)`.
+  Revision commands use `ExactRevisionLocatorV1(statblock_id, revision_id)`, not a
+  single-id `ResourceLocatorV1`. Tests use `FakeDefinitionProvider` plus
+  `InMemoryCandidateRepository`.
 - Provider outcomes are `success`, `refusal`, `incomplete`, `timeout`, `rate_limit`,
   and `failure`; application failures are prefixed `provider_`, with
-  `definition_invalid` reserved for parsing/semantic validation failures.
+  `definition_invalid`, `ruleset_mismatch`, `source_digest_mismatch`, and
+  resolver codes (`revision_not_found`, `statblock_not_found`, `source_unavailable`)
+  reserved for service-level failures.
+- Generation receipts store server-owned `caller_scope`/`actor` provenance and a
+  server-computed (optionally caller-verified) source description digest.
+  Warning-bearing candidate receipts may be persisted; invalid receipts are not.
+  `preserve_element_keys` is validated after revision (`ELEMENT_KEY_CHANGED` /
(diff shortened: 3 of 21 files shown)


Drakosfire (commented):
## Review verdict

Reviewed head `9c86043963363ffc9fdf63209f05e32a807516f5` against the checked-in PR16 handoff and the merged PR15 persistence contract.

The implementation gets several important boundaries right: the provider protocol is definition-only, the OpenAI adapter does not put server metadata into model output, the strict schema is loaded from the published artifact and fingerprinted, candidate-mode semantic validation occurs before persistence, asset failures do not destroy an otherwise valid candidate, and no HTTP routes are added.

The remaining defects affect the exact capability PR17 is expected to trust.

## Required changes

### 1. Rebase and retarget onto the final merged PR15 contract

PR15 merged at `38ed3dfec0ac354d0ced5c3e8fbdc23480cede0a` with reviewed head `54ed5e6e31de8f9b13bff1b77304d09fef7c8adb`. PR16 is still based on the earlier PR15 branch state at `f44aa883cc2edb01fd493c22fd3bd419d4e47f26`, is currently non-mergeable, and diverges from the final predecessor.

This is not only branch hygiene. The current PR16 copies the earlier repository/resource implementation, including the pre-final idempotency outcome shape and missing final immutable-command/read-lock fixes. A conflict resolution that favors this branch would regress PR15.

Required correction:

- rebase/retarget onto current `main` containing PR15 merge `38ed3df`;
- preserve all final PR15 repository, resource, error, timestamp, reconciliation, and immutable-command behavior;
- reapply only the PR16-owned generation/resource additions;
- update the PR16 and PR17 predecessor notes to describe the actual final PR15 signatures and semantics.

Focused proof:

- the rebased branch contains `IdempotencyOutcomeV1`, immutable persistence commands, locked in-memory reads, and final Firestore reconciliation tests;
- the complete PR15 suite still passes unchanged alongside PR16 tests.

### 2. Enforce the requested ruleset and make source/caller provenance truthful

The design requires the requested ruleset to constrain the Structured Outputs result. The compiler currently has no ruleset input and returns the generic schema, while the service never verifies that `definition.ruleset == command.ruleset`. A provider can therefore return a valid 2014 definition for a 2024 request and the service will persist it as a successful candidate.

The provenance fields are also not authoritative:

- `SourceSnapshotV1.description_digest` is caller-supplied and accepted without checking it against `description`;
- `command.caller` is never read or persisted anywhere;
- the receipt hard-codes the prompt version instead of using the prompt builder's `PROMPT_VERSION`, allowing prompt/receipt drift.

These permit a candidate receipt that claims the wrong input/ruleset or omits the caller provenance the handoff says must be stored.

Required correction:

- constrain the provider schema to the requested ruleset or reject any parsed definition whose ruleset differs exactly from the command;
- compute the source-description digest server-side and reject or ignore a mismatching supplied digest;
- persist typed caller scope/actor provenance in the generation receipt or another explicit server-owned candidate field;
- record the prompt version from the same versioned builder that produced the prompt.

Focused proof:

- a structurally valid wrong-edition provider payload is rejected and not persisted;
- supplied source-digest mismatch fails closed;
- generated and revised candidates retain exact caller provenance;
- changing `PROMPT_VERSION` changes the recorded receipt version through one source of truth.

### 3. Implement the promised revision invariant, not only revision prompt text

The checked-in handoff requires revision from either a submitted complete definition or an exact persisted revision, plus key-preservation validation. The current implementation does not establish that capability:

- `ResourceLocatorV1` carries only `resource_type` and one `resource_id`, but the final PR15 exact-read protocol requires both `statblock_id` and `revision_id`; no adapter connects `DefinitionResolver` to `RevisionRepository`;
- arbitrary resolver exceptions can escape the application boundary;
- `build_revision_prompt()` ignores `command.intent`, `command.context`, and optional source snapshot information;
- `preserve_element_keys` is only an instruction to the model—there is no comparison against the source definition and no warning when a retained conceptual element changes key;
- there is no revision service test at all.

The invalid state is a “successful revision” that silently ignores requested context/intent, cannot be resolved through the predecessor repository contract, or changes stable local references without a warning.

Required correction:

- define an exact revision locator compatible with PR15, carrying both logical statblock and revision identity, and provide a repository-backed resolver;
- map not-found/unavailable resolver outcomes into stable application failures rather than leaking exceptions;
- include revision intent/context/source constraints in the prompt;
- compare source and output rule elements and preserve keys or emit deterministic validation warnings when conceptual elements are retained under changed keys;
- retain the exact source revision locator and/or source-definition digest in the generation receipt.

Focused proof:

- revision works from a submitted definition and from a real in-memory `RevisionRepository` exact read;
- missing statblock, missing revision, wrong locator type, and unavailable repository map to typed outcomes;
- revision intent/context appear in the provider prompt;
- retained elements keep keys, genuine replacements may change keys, and suspicious changes produce stable warnings;
- the source accepted revision remains immutable and only a new candidate is created.

### 4. Close provider, configuration, and asset partial-outcome boundaries

`GenerationServiceV1` calls `provider.generate_definition()` without a guard. A provider implementation or fake callback that raises therefore leaks a raw exception despite the handoff's explicit provider-failure boundary.

The default model policy lookup is also broken: from `statblocks_v1/application/settings.py`, `Path(__file__).resolve().parents[3]` resolves one directory above the repository/package root. `from_environment()` therefore cannot reliably find the claimed repository `MODEL_POLICY.json` in a checkout or installed package, and ordinary tests do not exercise this path.

Finally, `generate_images=True` with no configured asset generator silently returns no assets and no warning. The PR adds a durable free-form `asset_warnings: list[str]` field, but the active design and PR17 response notes do not define that contract, and a string list is not the typed partial outcome requested by the handoff.

Required correction:

- catch unexpected provider exceptions at the application boundary and return the typed provider-failure outcome;
- resolve model policy through a real repository/package configuration mechanism and validate timeout/retry/TTL values;
- when images are requested but unavailable or fail, return a deterministic typed partial warning;
- reconcile the candidate warning shape with the active design and update PR17 if it is intentionally durable/public.

Focused proof:

- a provider that raises cannot leak its exception;
- default settings load in an ordinary offline test from the actual tracked policy source;
- malformed/negative environment values fail as typed configuration errors;
- requested image generation with a missing generator and with a throwing generator both preserve the candidate and expose stable typed warnings.

### 5. Supply the mandatory boundary proof

The current generation-service test file covers one valid fixture, one refusal, one invalid reference, and one asset exception. It does not prove most of the mandatory handoff matrix. The opt-in OpenAI test only constructs the provider; it never performs a Structured Outputs call.

Before merge, add offline proof for:

- simple and advanced/mythic valid fixtures;
- persisted **and retrievable** candidate state;
- deterministic receipt/source/caller fields;
- refusal, incomplete, timeout, rate limit, generic provider failure, raised provider exception, and malformed payload;
- warning-bearing candidates;
- both revision source modes and key-preservation behavior;
- strict-schema compatibility for representative discriminated mechanic/effect unions;
- server metadata rejection from provider payloads.

Keep real calls opt-in, but implement the handoff's actual live smoke tests: one simple generation and one advanced generation whose returned content validates as the canonical definition and contains no outer server envelope.

## Non-blocking observations

- `GenerationFailureV1.kind` is currently an arbitrary string even though PR17 will map it as a stable outcome catalog. A typed enum/discriminated model would make that successor boundary safer.
- `latency_ms=outcome.latency_ms or measured` replaces a valid reported value of `0`; use an explicit `is not None` check if zero must be preserved.

No GitHub status checks or workflow runs are attached to this head, so the PR body's `63 passed, 2 skipped` result is author-reported rather than independently visible here.

Drakosfire:
Addressed review 4730007204 on the rebased tip:

1. **Rebased onto main** (merged PR15 at `38ed3df` / tip includes `54ed5e6`) and retargeted base to `main`.
2. **Ruleset match** — generated `definition.ruleset` must equal the requested ruleset (`ruleset_mismatch`).
3. **Trustworthy provenance** — receipts store server-owned `caller_scope`/`actor`; source digests are server-computed and caller-supplied digests are verified (`source_digest_mismatch`).
4. **Revision support** — `ExactRevisionLocatorV1(statblock_id, revision_id)` + `PersistenceDefinitionResolver` using PR15 `get_revision`; intent/context in revision prompts; resolver/`StatblockV1Error` mapped to typed failures; `preserve_element_keys` validated (`ELEMENT_KEY_CHANGED` / `ELEMENT_KEY_DROPPED`).
5. **Failure/config boundaries** — provider exceptions → `provider_failure`; in-repo `MODEL_POLICY.json` (no path above the repo root); requested images without a generator yield an explicit asset warning.
6. **Test matrix** — revision, warning-bearing candidates, full provider outcomes, advanced fixture generation, settings fail-closed, and opt-in live OpenAI generate calls.

Offline lane: `134 passed, 6 skipped`.

Drakosfire (commented):
## Re-review verdict

**BLOCKING — substantially improved, but four trust-boundary defects remain.**

Reviewed head `7046fe14911da3c4f68d236b46856bee0b23bbbd` against prior review `4730007204`, the checked-in PR16 handoff, and merged PR15 at `38ed3dfec0ac354d0ced5c3e8fbdc23480cede0a`.

## Prior finding ledger

- **Rebase/final PR15 contract — closed.** PR16 now targets `main`, is exactly ahead of merged PR15, is mergeable, and no longer replaces final repository implementations.
- **Ruleset/source/caller truth — mostly closed.** Wrong output rulesets are rejected, source-description digests are server-computed and verified, caller scope/actor are persisted, and prompt version comes from the prompt builder constant.
- **Revision integration — materially improved.** Exact `(statblock_id, revision_id)` locators, a PR15-backed resolver, typed resolver failures, and revision intent/context prompt coverage are present.
- **Provider/live-test boundary — mostly closed.** Unexpected provider exceptions become `provider_failure`, the offline outcome matrix is broader, and the opt-in tests now make real Structured Outputs calls.

The following issues remain blocking because PR17 is explicitly told to expose these candidate and receipt shapes.

### 1. Inline revision sources still have no exact mechanics provenance

For revision from an exact persisted locator, the receipt records `source_locator`. For revision from `source_definition`, however, `source_locator` is `None`, and the service records only the optional authored **description** digest. It never records the PR14 definition digest of the actual mechanics object embedded in the revision prompt.

This means two different source definitions can produce candidates with indistinguishable source provenance whenever the same request/caller/description metadata is used. The receipt cannot establish which exact mechanics were revised.

Required correction:

- add a server-owned `source_definition_digest` to the generation receipt;
- compute it with the final PR14 `compute_definition_digest(source_definition)` for every revision;
- for locator-based revision, either have the resolver return the exact revision resource/material or independently verify/store the resolved definition digest alongside the locator;
- keep `source_description_digest` separate—it describes authored prose, not source mechanics.

Focused proof:

- two different inline source definitions produce different recorded source-definition digests;
- locator revision records both the exact locator and the digest of the resolved immutable revision;
- caller-supplied description metadata cannot substitute for or alter the mechanics digest.

### 2. The key-preservation validator misses key repurposing

`_with_key_preservation_warnings()` suppresses `ELEMENT_KEY_DROPPED` whenever the original key appears anywhere in the revised definition. It does not verify that the key still identifies the same conceptual element.

For example:

```text
source: Greatclub action, key = greatclub
revised: Fire Breath action, key = greatclub
```

There is no `(section, name)` match, but `greatclub` remains in `revised_keys`, so the code emits no warning. This is the most dangerous failure mode: a stable local identity is silently reassigned to unrelated mechanics, allowing downstream references to appear preserved while changing meaning.

The identity dictionary also collapses multiple legitimate elements with the same normalized section/name onto one entry, making warnings ambiguous.

Required correction:

- compare both directions: key continuity and conceptual identity continuity;
- emit a stable `ELEMENT_KEY_REPURPOSED` warning when an existing key now points to a materially different element;
- handle duplicate display identities as an ambiguous match rather than dict-overwriting one element;
- keep these revision-specific findings in a versioned validation pass/receipt so `validator_version` truthfully identifies the producer of the warning catalog.

Focused proof:

- same concept + changed key warns;
- removed concept + removed key warns;
- genuinely new concept reusing an old key warns as repurposed;
- duplicate same-name actions do not produce nondeterministic or misattributed warnings;
- unchanged keys/concepts remain warning-free.

### 3. Generation settings still do not fail closed

The policy path is fixed, but `GenerationSettingsV1` accepts invalid values and `from_environment()` leaks raw parsing exceptions.

Current examples:

- `STATBLOCKS_V1_OPENAI_TIMEOUT_SECONDS=garbage` raises raw `ValueError`, not `InternalServiceMisconfiguredError`;
- zero/negative timeout and negative retries are accepted and passed to the SDK;
- zero/negative candidate TTL is accepted, allowing the service to return a successful candidate that is already expired;
- direct dataclass construction bypasses any environment-only validation.

Required correction:

- validate at the settings model itself, not only in `from_environment()`;
- require a non-empty model, finite positive timeout, retries `>= 0`, and TTL `> 0`;
- translate malformed environment input into `InternalServiceMisconfiguredError` with safe field-level details;
- add focused tests for malformed, zero, negative, NaN/infinite, and direct-construction cases.

### 4. Asset partial outcomes are still untyped and can lose the actual generation brief

The handoff requires typed partial outcomes/warnings. The candidate currently exposes `asset_warnings: list[str]`, and tests assert substrings from human prose. PR17 would therefore inherit an unstable, typo-sensitive public warning contract with no machine-readable code.

There is also a direct provenance gap when `generate_images=True` and `include_generation_brief=False`: the asset generator receives a synthesized brief using `definition.identity.name`, but `candidate.asset_brief` remains `None`. A generated asset can therefore be persisted without recording the brief that produced it.

Required correction:

- replace free-form warning strings with a typed warning model/code catalog, such as `asset_generator_unconfigured` and `asset_generation_failed`;
- construct the effective `AssetBriefV1` once, pass that exact object to the generator, and persist that exact brief whenever asset generation is attempted;
- define or reject the `generate_images=True` / `include_generation_brief=False` combination rather than silently generating from an unrecorded fallback;
- update the active contract design and PR17 response notes for the intentionally public warning shape.

Focused proof:

- missing and throwing generators yield distinct stable codes;
- generated assets always retain the exact effective brief used;
- no-images/no-brief remains clean;
- the candidate round-trips through both in-memory and Firestore repositories with typed warnings intact.

## Verification note

GitHub reports the PR mergeable and exactly based on merged PR15. No status checks or workflow runs are attached to this head. I could not independently execute the suite because this review environment still cannot resolve GitHub for a checkout, so the test-plan result is not independently verified here.

## Non-blocking observations

- `GenerationFailureV1.kind` remains an arbitrary string even though PR17 treats it as a stable catalog. An enum/discriminated model would make the HTTP mapping safer.
- `latency_ms=outcome.latency_ms or measured` replaces a legitimate provider-reported value of `0`; use an explicit `is not None` check.
- A root-level `MODEL_POLICY.json` works in this repository checkout but is not automatically included in a normal wheel by the current package configuration. Confirm the production deployment always includes the repository root, or move policy loading to a packaged resource/configuration source.

Drakosfire (commented):
## Addressed review 4730113665 (head `a0d7561`)

All four remaining trust-boundary blockers are closed:

1. **Inline revision provenance** — `GenerationReceiptV1.source_definition_digest` is always set for revisions via `compute_definition_digest(source)`. Locator revisions also retain `source_locator`. Description digests remain separate and cannot substitute for mechanics identity. Tests cover distinct inline digests and locator+digest parity with the immutable revision.

2. **Key repurposing** — preservation pass now emits `ELEMENT_KEY_REPURPOSED` when a retained key maps to a different conceptual element, `ELEMENT_KEY_IDENTITY_AMBIGUOUS` for non-unique section/name matches (no misattributed CHANGED), plus existing CHANGED/DROPPED. Receipt `validator_version` appends `+statblock-key-preservation-v1`.

3. **Fail-closed settings** — `GenerationSettingsV1.__post_init__` requires non-empty model, finite timeout `> 0`, retries `>= 0`, TTL `> 0`. Malformed env values raise `InternalServiceMisconfiguredError` instead of raw `ValueError`.

4. **Typed asset outcomes** — `asset_warnings` is now `list[AssetWarningV1]` with codes `asset_generator_unconfigured` / `asset_generation_failed`. When `generate_images=True`, the exact effective `AssetBriefV1` (description brief or name fallback) is passed to the generator and persisted on the candidate.

PR17 predecessor notes updated. Isolated lane green.

Drakosfire (commented):
## Re-review verdict

**BLOCKING — the prior four findings are closed, but three successor-facing gaps remain.**

Reviewed head `a0d7561d1fda5fc58f08aed64924023e7536d762` against prior review `4730113665`, the complete PR16 generation/revision invariant, and merged PR15.

## Prior finding ledger

1. **Exact revision mechanics provenance — closed.** Every revision now records `source_definition_digest`; locator revision retains both locator and digest.
2. **Key repurposing — closed for the implemented identity policy.** The versioned pass now handles changed, dropped, repurposed, and ambiguous keys.
3. **Fail-closed settings — closed.** Direct and environment construction reject malformed, nonfinite, zero, and negative values with `InternalServiceMisconfiguredError`.
4. **Typed asset partial outcomes — closed in the application/in-memory path.** Stable codes are present, provider-reported latency `0` is preserved, and the exact effective brief is stored.

The remaining issues were found by rechecking how this service will be invoked and exposed by PR17.

### 1. Generation/revision commands are mutable across the long-running provider call

All command and nested intent models inherit `StrictModel`, whose configuration only forbids extra fields; it does not freeze or defensively copy instances. Pydantic retains nested model instances, so `source_definition`, `source_locator`, `ruleset`, and `caller` remain mutable references owned by the caller.

`revise()` then:

1. resolves/builds the prompt and computes `source_definition_digest` from the current source;
2. calls the potentially long-running provider;
3. after the provider returns, reuses the same source reference for key-preservation validation and the same locator/caller/ruleset references for output validation and receipt construction.

A concurrent mutation can therefore create contradictory durable truth. Examples:

- resolve `(sb_A, rev_A)`, then mutate the locator during the provider call; the candidate receipt can claim `(sb_B, rev_B)` while its prompt/digest came from revision A;
- compute the source-definition digest from source A, mutate that definition while the provider is running, then run key-preservation warnings against source B;
- mutate the requested ruleset or caller while the provider is running, changing acceptance or persisted provenance after the request was issued.

This is the same immutable-operation boundary PR15 had to establish for persistence commands.

Required correction:

- create one deep, immutable command snapshot at `generate()` / `revise()` entry;
- derive prompt, verified description digest, source definition digest, locator, caller, ruleset, asset options, and preservation inputs from that snapshot exactly once;
- never retain caller-owned mutable objects across the provider invocation;
- alternatively make the command tree genuinely frozen and prove nested collections/models cannot mutate.

Focused proof:

- use a barrier-controlled provider and mutate the original command/source/locator while it is blocked;
- the recorded locator, caller, ruleset, source digest, key-preservation comparison, and asset intent must all remain identical to the operation snapshot;
- mutating the returned candidate must not alter the stored candidate (retain PR15 repository isolation).

### 2. The active contract design is still stale relative to the public candidate shape

The implementation and PR17 handoff now make these durable/public:

- `GenerationReceiptV1.caller_scope` / `actor`;
- `source_description_digest` and `source_definition_digest`;
- exact revision locator provenance;
- typed `AssetWarningV1` values and warning codes;
- the rule that an attempted asset generation stores the exact effective brief.

However, `Docs/Design/DESIGN-dungeonbuddy-statblock-contract-v1.md` still lists `GeneratedStatblockCandidateV1` with only `asset_brief` and `assets`; it does not include `asset_warnings` or define these receipt/provenance obligations. The previous review explicitly required updating the active contract design, not only the successor handoff.

Required correction:

- update the authoritative design’s candidate and generation-receipt sections with the final fields, ownership, warning-code catalog, and brief/provenance semantics;
- keep the PR16 and PR17 handoffs aligned with that source of truth.

### 3. Firestore proof for the new nested contract is still missing

The previous review required typed warnings to round-trip through both in-memory and Firestore repositories. The new test proves only `InMemoryCandidateRepository` round-trip.

The existing Firestore emulator candidate test constructs a minimal candidate without:

- `GenerationReceiptV1.source_definition_digest`;
- exact source locator;
- typed `AssetWarningV1` / enum values;
- effective asset brief.

The serializer appears designed to encode nested enums, but this public persistence shape is not yet proven against the real adapter/emulator.

Required correction:

- add an emulator-gated candidate create/get round-trip containing a complete generation receipt, exact locator, source-definition digest, effective asset brief, and both typed asset-warning codes;
- assert raw Firestore values are stable strings/timestamps where appropriate and `GeneratedStatblockCandidateV1.model_validate()` restores the typed models exactly;
- use unique test IDs or cleanup so repeated emulator runs remain deterministic.

## Verification note

GitHub reports the PR mergeable and based on merged PR15. No status checks or workflow runs are attached to this head. I also could not independently run the suite because this environment still cannot resolve GitHub for a checkout, so the author-reported isolated-lane result is not independently verified here.

## Non-blocking observations

- `GenerationFailureV1.kind` remains a free-form string even though PR17 treats it as a stable HTTP-facing catalog; an enum/discriminated result remains safer.
- Root-level `MODEL_POLICY.json` is not automatically included by the current wheel packaging configuration; confirm production deploys from a source tree or move policy data into a packaged resource.
- Consider re-exporting the new locator/receipt/asset warning types from `statblocks_v1.domain` if that module is intended as the public domain import surface.

Because this is a self-owned PR, this is submitted as a blocking `COMMENT` review rather than `REQUEST_CHANGES`.

Drakosfire (commented):
## Addressed review 4731513393 (head updated)

All three successor-facing gaps are closed:

1. **Pinned operation intent** — `generate()` / `revise()` deep-copy the command and build a frozen `_PinnedOperationIntent` (prompt, digests, ruleset, caller, locator, source definition, asset options) before the provider call. A barrier-controlled mutation test mutates the original command while the provider is blocked; receipt locator/caller/digests/asset intent remain the pinned values. Mutating the returned candidate does not alter the stored one.

2. **Active contract design** — `DESIGN-dungeonbuddy-statblock-contract-v1.md` §4.2 / §10.3 now document `asset_warnings`, receipt provenance (`caller_scope`/`actor`, description + definition digests, exact locator), and the exact-effective-brief rule. PR17 predecessor notes mention the pin-before-provider boundary.

3. **Firestore typed round-trip** — emulator-gated `test_firestore_candidate_typed_contract_round_trip` persists a complete candidate (generation receipt, locator, source-definition digest, brief, both warning codes), asserts raw Firestore enum/string/timestamp shapes, and `model_validate` equality after get. Uses unique candidate IDs.

Isolated lane green (Firestore cases remain emulator-gated skips without `FIRESTORE_EMULATOR_HOST`).

Drakosfire (commented):
## Re-review verdict

Reviewed head `c5de67420a9f46454075a4bd23bf24c8dbf71b5b` against prior blocking review `4731513393`, the checked-in PR16 handoff, merged PR15, and the successor-facing PR17 notes.

## Prior finding ledger

### 1. Mutable operation intent — closed

Generate and revise commands are deep-copied before any provider call. The resulting frozen `_PinnedOperationIntent` separately deep-copies the ruleset, caller, exact locator, and source definition, while prompt text, source digests, asset options, and key-preservation policy are derived from that snapshot.

The focused revision test blocks inside the provider, mutates every caller-owned field, and proves the candidate receipt, source-definition digest, locator, asset brief, warning outcome, and key-preservation pass remain anchored to the original operation. It also verifies the repository retains its own immutable copy after the returned candidate is mutated.

### 2. Authoritative contract design — closed

`DESIGN-dungeonbuddy-statblock-contract-v1.md` now documents the actual candidate envelope and server-owned generation receipt, including:

- caller scope and actor provenance;
- description and source-definition digests;
- exact revision locator semantics;
- typed `AssetWarningV1` outcomes;
- created/expires timestamps;
- the exact effective asset brief used when image generation is attempted.

The stale candidate IDs, contract/version values, and receipt example were also corrected.

### 3. Firestore typed round trip — closed

The emulator suite now persists and reloads a candidate containing the complete new public surface: nested generation receipt timestamps, description/mechanics digests, exact locator, asset brief, assets, and typed asset warnings. It verifies native Firestore timestamps and complete `model_dump(mode="json")` equality after repository retrieval.

## Full invariant pass

No regression found in the broader PR16 contract:

- PR16 remains exactly based on merged PR15 and is mergeable;
- the provider emits only `StatblockDefinitionV1`;
- output ruleset must match the pinned request;
- source prose digests are verified and source mechanics digests are server-computed;
- revision supports inline definitions and exact immutable revision reads;
- key changes, drops, repurposing, and ambiguous conceptual identities produce versioned warnings;
- unexpected provider and resolver failures map to stable application outcomes;
- invalid provider definitions never become candidates;
- settings fail closed;
- asset failures preserve valid candidates with typed partial outcomes;
- prompt/schema/provider provenance remains server-owned;
- no HTTP routes are introduced in PR16.

## Verification note

GitHub reports the PR mergeable. No status checks or workflow runs are attached to this head. The Firestore proof remains emulator-gated, and I could not independently execute the suite because this review environment cannot resolve GitHub for a checkout, so the branch's reported green isolated lane is not independently reproduced here.

## Non-blocking hardening

The command models themselves remain assignment-mutable and `model_copy(deep=True)` does not revalidate a command that was deliberately mutated into an invalid state *before* service entry. The documented write-once route-boundary rule and immediate snapshot make this acceptable for PR16, but freezing commands or revalidating the snapshot would further harden reuse from less-controlled callers.

Because this is a self-owned PR, I am submitting a merge-ready `COMMENT` review rather than `APPROVE`.

PR #17, opened 2026-07-18
feat(statblocks_v1): PR17 candidate generation and validation API

Drakosfire:
## Summary
- Rebases PR17 onto merged PR16/`main` as a focused API slice (no obsolete PR13–PR16 reintroductions).
- Adds authenticated generate/revise/validate/candidate-read routes with `ExactRevisionLocatorV1` and production `PersistenceDefinitionResolver` wiring.
- Extends the scoped `StatblockV1HTTPError` seam (path-scoped request validation) and maps the full final PR16 failure catalog; unknown kinds fail closed as `generation_failed`.
- **Defers candidate idempotency** so PR15 statblock/revision idempotency outcomes are not weakened; `request_id` remains receipt correlation only.

## Test plan
- [x] `./scripts/run_statblocks_v1_tests.sh -q`
- [x] Exact-revision revise + inline revise
- [x] Invalid source combinations → 422 `invalid_request`
- [x] Full provider/ruleset/digest failure catalog
- [x] OpenAPI declares `ErrorEnvelopeV1` on failure statuses
- [x] Legacy routes keep FastAPI `{"detail": ...}` validation shape
- [x] Production `build_generation_service` wires persistence resolver

### DIFF
--- a/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
+++ b/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
@@ -1,6 +1,7 @@
 # HANDOFF — PR17 Statblock v1 candidate API
 
-**Status:** READY AFTER PR15 AND PR16  
+**Status:** IN REVIEW — rebased onto merged PR16; candidate idempotency deferred  
+
 **Target repository:** `Drakosfire/DungeonMindServer`  
 **Predecessors:** repository persistence and generation service  
 **Successor:** `HANDOFF-pr18-statblock-v1-revision-resource-api.md`
--- a/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
+++ b/Docs/Plans/HANDOFF-pr17-statblock-v1-candidate-api.md
@@ -16,8 +16,9 @@
 - Candidates expire at `expires_at` (also the Firestore TTL field). Collections are
   `dungeonbuddy_statblock_candidates_v1`, `dungeonbuddy_statblocks_v1/revisions`,
   and `dungeonbuddy_statblock_idempotency_v1`.
-- Reuse `compute_request_digest(operation, payload)` for request replay: it includes
-  operation parameters rather than using only the mechanics definition digest.
+- PR15 `compute_request_digest(operation, payload)` and idempotency records apply to
+  logical-statblock create/append only. Candidate generate/revise do **not** use
+  that path in PR17 (see §9).
 
 ## PR16 predecessor completion notes
 
@@ -188,15 +189,19 @@ Route tests must replace all external dependencies with fake or in-memory implem
 
 ## 9. Idempotency and retries
 
-Generation and revision requests should accept stable request IDs. Decide and document whether they are idempotent operations.
+**Decision (PR17): candidate idempotency is deferred.**
 
-Recommended:
+Generation and revision accept `request_id` as correlation / generation-receipt
+metadata only. They are **not** idempotent operations in this PR:
 
-- same caller + request ID + same request returns the same candidate;
-- same request ID with different request body returns conflict;
-- provider invocation is not repeated after a successful stored result.
+- identical `request_id` values may produce distinct candidates;
+- there is no candidate-level replay, conflict, or provider-once reservation;
+- PR15 create/append idempotency (`IdempotencyOutcomeV1` for
+  `statblock_id`/`revision_id`) is unchanged and must not be overloaded for
+  candidates.
 
-Use the PR15 idempotency service if implemented for candidate commands; do not invent a separate route-local cache.
+A later PR may add a candidate-specific durable reservation protocol; until then
+callers must not assume request-id replay.
 
 ## 10. Suggested files
 
@@ -218,16 +223,20 @@ Required route tests:
 
 - health capabilities;
 - generate success with fake provider;
-- revise success from definition and exact revision source;
+- revise success from definition and exact revision source (inline + locator);
+- invalid revision source combinations (neither / both) → typed 422;
 - validate success/warnings/failure;
 - candidate exact read;
 - candidate expired/not found;
 - missing/wrong/correct authentication;
-- request ID replay;
-- conflicting request ID reuse;
-- provider refusal, timeout, and malformed output mapping;
+- distinct generates with different `request_id` yield distinct candidates
+  (idempotency deferred — no replay/conflict claims);
+- provider refusal, timeout, malformed output, ruleset mismatch, and
+  source-digest mismatch mapping (not false `provider_unavailable`);
+- missing revision / missing statblock typed codes;
+- legacy routes retain FastAPI `{"detail": ...}` validation envelopes;
 - no Markdown or `combat_defaults` in response;
-- OpenAPI paths and response models exist.
+- OpenAPI paths and `ErrorEnvelopeV1` failure response models exist.
 
 No route test may require OpenAI or Firestore credentials.
 
@@ -254,11 +263,12 @@ PR17 is complete when:
 
 ## 14. Successor handoff
 
-Before merge, update PR18 with:
+PR18 predecessor notes already record:
 
 - final route prefix and auth dependency;
-- API error envelope and status policy;
-- request ID/idempotency behavior;
-- candidate lookup/service signatures;
-- OpenAPI path names;
-- route-test app factory and dependency overrides.
+- path-scoped API error envelope and status policy;
+- `request_id` as correlation only (candidate idempotency deferred);
+- candidate lookup/service signatures and exact-revision wiring;
+- OpenAPI operation IDs and `ErrorEnvelopeV1` failure models;
+- route-test dependency overrides (`get_generation_service`,
+  `get_candidate_repository`, `get_clock`).
--- a/Docs/Plans/HANDOFF-pr18-statblock-v1-revision-resource-api.md
+++ b/Docs/Plans/HANDOFF-pr18-statblock-v1-revision-resource-api.md
@@ -15,8 +15,9 @@
   keep FastAPI `{"detail": ...}`.
 - Generation failure kinds from final PR16 map explicitly (`ruleset_mismatch`,
   `source_digest_mismatch`, `invalid_request`, `revision_not_found`,
-  `statblock_not_found`, provider outcomes). Unknown kinds fail closed as
-  `500 generation_failed`, never a false `provider_unavailable`.
+  `statblock_not_found`, `persistence_unavailable` → 503, provider outcomes).
+  Unknown kinds fail closed as `500 generation_failed`, never a false
+  `provider_unavailable`.
 - Revision uses `ExactRevisionLocatorV1(statblock_id, revision_id)`. Production
   wiring in `app.py` injects `PersistenceDefinitionResolver` over
   `FirestoreStatblockPersistenceRepository`. Exactly one of `source_definition` /
--- a/Docs/Plans/HANDOFF-pr18-statblock-v1-revision-resource-api.md
+++ b/Docs/Plans/HANDOFF-pr18-statblock-v1-revision-resource-api.md
@@ -5,6 +5,36 @@
 **Predecessors:** PR14 trust core, PR15 repositories, PR17 router/error boundary  
 **Successor:** `HANDOFF-pr19-statblock-v1-assets-openapi-consumer-contract.md`
 
+## PR17 predecessor completion notes
+
+- The v1 route prefix is `/api/internal/dungeonbuddy/v1`; all routes use
+  `require_internal_service_auth` and the `X-DungeonBuddy-Internal-Key` shared key.
+- Errors use top-level `{ "error": { "code", "message", "details"? } }` via the
+  existing `StatblockV1HTTPError` / `register_error_handlers` seam. Request
+  validation is path-scoped: only v1 paths get the typed envelope; legacy routes
+  keep FastAPI `{"detail": ...}`.
+- Generation failure kinds from final PR16 map explicitly (`ruleset_mismatch`,
+  `source_digest_mismatch`, `invalid_request`, `revision_not_found`,
+  `statblock_not_found`, provider outcomes). Unknown kinds fail closed as
+  `500 generation_failed`, never a false `provider_unavailable`.
+- Revision uses `ExactRevisionLocatorV1(statblock_id, revision_id)`. Production
+  wiring in `app.py` injects `PersistenceDefinitionResolver` over
+  `FirestoreStatblockPersistenceRepository`. Exactly one of `source_definition` /
+  `source_locator` is enforced on the request DTO (422 `invalid_request`).
+- Candidate generate/revise treat `request_id` as correlation/receipt metadata
+  only. **Candidate idempotency is deferred** (not implemented) so PR15
+  statblock/revision idempotency outcomes remain untouched. Do not assume
+  request-id replay for candidates.
+- Candidate lookup is `CandidateRepository.get(candidate_id, now=...)`; async
+  routes call synchronous repositories and generation through `asyncio.to_thread`.
+- OpenAPI operations: `generate_statblock_candidate_v1`,
+  `revise_statblock_candidate_v1`, `validate_statblock_definition_v1`,
+  `get_statblock_candidate_v1`. ErrorEnvelopeV1 is declared on failure statuses.
+  No accepted-resource routes exist yet.
+- Tests override `get_generation_service`, `get_candidate_repository`, and
+  `get_clock` with `FakeDefinitionProvider`, `InMemoryCandidateRepository` /
+  `InMemoryStatblockPersistenceRepository`, and deterministic clocks.
+
 ## 0. Mission
 
 Expose authoritative logical-statblock and immutable-revision resources through the DungeonBuddy v1 router.
--- a/app.py
+++ b/app.py
@@ -55,8 +55,15 @@ from routers.cardgenerator_compatibility_router import router as cardgenerator_c
 from routers.statblockgenerator_router import router as statblockgenerator_router
 
 # Import DungeonBuddy statblock v1 bounded-context router
+from firestore.firebase_config import db as dungeonbuddy_statblocks_v1_db
+from statblocks_v1.api import dependencies as dungeonbuddy_statblocks_v1_dependencies
 from statblocks_v1.api.http_errors import register_error_handlers as register_statblocks_v1_error_handlers
 from statblocks_v1.api.router import router as dungeonbuddy_statblocks_v1_router
+from statblocks_v1.infrastructure.runtime import (
+    build_candidate_repository as build_statblocks_v1_candidate_repository,
+    build_generation_service as build_statblocks_v1_generation_service,
+    build_persistence_repository as build_statblocks_v1_persistence_repository,
+)
 
 # Import PlayerCharacterGenerator router
 from routers.playercharactergenerator_router import router as playercharactergenerator_router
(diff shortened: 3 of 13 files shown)


Drakosfire (commented):
## Review verdict

Reviewed head `0817e9883be996669725128b47e2eeae926630d1` against the checked-in PR17 handoff, merged PR16 head `c5de67420a9f46454075a4bd23bf24c8dbf71b5b`, and PR16 merge commit `5516ede2ad44d288defa453239aabcd3689bf943`.

The route surface is pointed in the right direction: the four required paths exist, auth is applied at router level, blocking service/repository calls are offloaded with `asyncio.to_thread`, validation remains non-persisting, candidate reads are exact, health advertises only candidate capabilities, and no accepted-resource routes are added.

The following boundary failures must be closed before PR18 can safely consume this API.

### 1. Rebase onto the final merged PR16 contract and discard stale predecessor implementations

PR17 still targets `feat/statblocks-v1-pr16-generation` at the intermediate SHA `7046fe1`, is marked non-mergeable, and its head diverges from final PR16 by 25 predecessor commits. The current 71-file diff reintroduces old PR13–PR16 implementations rather than presenting a PR17-only API slice.

This is materially unsafe, not cosmetic. The branch currently carries obsolete versions of:

- revision locators (`ResourceLocatorV1` instead of final `ExactRevisionLocatorV1(statblock_id, revision_id)`);
- `GenerationServiceV1` without the final pinned-intent, ruleset, source-digest, caller-provenance, key-preservation, and typed-asset contracts;
- settings and provider failure boundaries;
- PR15 repository/idempotency resource shapes and reconciliation behavior;
- the API auth import boundary that final main intentionally isolates from `routers.__init__` side effects.

Required correction:

- rebase/retarget onto current `main` containing merge `5516ede`;
- preserve final PR15/PR16 implementations and tests exactly;
- reapply only PR17-owned DTO, route, runtime-wiring, error-transport, route-test, capability, and successor-handoff changes;
- do not resolve conflicts by taking this branch's predecessor files;
- update the PR body and PR18 notes to the final signatures.

The resulting diff should be a focused PR17 slice, not a re-addition of the bounded context.

### 2. Make persisted-revision revision work in production and fail invalid source requests as 422

The handoff requires revision from either a complete submitted definition or an exact persisted revision. The current HTTP/runtime slice does not provide that capability:

- `ReviseCandidateRequestV1.source_locator` uses generic `ResourceLocatorV1`, incompatible with final PR16's `ExactRevisionLocatorV1`;
- `build_generation_service()` never injects `PersistenceDefinitionResolver` or a `StatblockPersistenceRepository`, so a production locator request can only return `source_unavailable`;
- the route test covers only inline `source_definition`, not an actual repository-backed exact revision;
- the request DTO does not enforce exactly one of `source_definition` / `source_locator`. Final command construction performs that validation inside the endpoint, where a Pydantic `ValidationError` is not a FastAPI request-parse error and can become a 500.

Required correction:

- use `ExactRevisionLocatorV1(statblock_id, revision_id)` at the wire boundary;
- construct/inject the final Firestore persistence repository and `PersistenceDefinitionResolver` through dependencies;
- validate the mutually exclusive source fields at the DTO boundary or explicitly map command-construction validation to `invalid_request` / 422;
- preserve final PR16 source locator and source-definition digest provenance.

Focused proof:

- revise succeeds from an inline definition;
- revise succeeds from a real in-memory exact revision read;
- production dependency construction includes the resolver path;
- neither source and both sources return the typed 422 envelope;
- missing statblock, missing revision, and persistence-unavailable paths retain distinct stable codes/statuses.

### 3. Extend the existing scoped v1 error transport; do not replace global FastAPI behavior

Merged main already has `StatblockV1HTTPError`, a v1-specific handler, and typed router-level auth responses. PR17 instead registers handlers for global `RequestValidationError`, `HTTPException`, and `StatblockV1Error` on the production app. That changes validation and HTTP exception behavior for every legacy router in DungeonMindServer, directly violating the PR17 non-goal of no legacy endpoint changes.

The generation failure mapping is also incomplete for final PR16. `ruleset_mismatch`, `source_digest_mismatch`, `invalid_request`, `revision_not_found`, and `statblock_not_found` currently fall through to `503 provider_unavailable`, converting client/source errors into a false provider outage.

Finally, the new route decorators declare only the success model. FastAPI will document its default validation schema while runtime returns `ErrorEnvelopeV1`, and provider/not-found/expired responses are absent from OpenAPI. The route test checks path existence only, not the required response models.

Required correction:

- extend the existing `statblocks_v1.api.http_errors` seam rather than globally handling FastAPI exception classes;
- preserve non-v1 exception behavior unchanged;
- map the complete final PR16 outcome catalog explicitly and fail closed on unknown codes without mislabeling them as provider failures;
- normalize request-validation details into JSON-safe data before constructing the envelope;
- declare typed `ErrorEnvelopeV1` responses for auth and each operation-specific status in OpenAPI.

Focused proof:

- all final generation/resolver outcomes map to the intended status and stable code;
- malformed and cross-field-invalid requests return a serializable 422 envelope;
- OpenAPI references `ErrorEnvelopeV1` for documented failure responses;
- an unrelated legacy route retains its prior FastAPI validation/HTTPException behavior.

### 4. Redesign or explicitly defer candidate idempotency; the current workflow is neither compatible nor atomic

PR17 promises same-request replay and one provider invocation, but implements:

```text
read idempotency
→ call provider
→ create candidate
→ create idempotency record
```

That sequence is not atomic. Concurrent identical requests can both invoke the provider and create different candidates before racing to record the key. A process failure after candidate creation but before idempotency recording also causes the retry to invoke the provider again and leaves an orphan candidate.

The implementation additionally conflicts with final PR15:

- final `CandidateRepository` is `create/get` only;
- final `IdempotencyRecordV1.outcome` is the exact statblock/revision `IdempotencyOutcomeV1`, not a generic candidate `ResourceLocatorV1`;
- taking this branch's generic outcome would regress the reviewed create/append replay invariant;
- `actor` is excluded from the request digest while every request uses caller scope `dungeonbuddy`, so two asserted actors can collide on one request ID and replay a candidate carrying the first actor's provenance.

The PR17 handoff says to **decide and document** idempotency. Either choice is valid, but the current partial guarantee is not.

Required correction, choose one:

1. **Defer candidate idempotency:** remove replay/conflict claims, repository extensions, tests, and PR18 completion notes; or
2. **Implement it as a durable application boundary:** add a candidate-specific typed outcome and atomic reservation/commit/reconciliation protocol without weakening final PR15's statblock/revision outcome. Define whether delegated actor identity participates in scope/digest and how expired candidates replay.

Focused proof if retained:

- concurrent identical requests invoke the provider once and yield one candidate;
- same key/different intent conflicts;
- the candidate write/idempotency write failure window reconciles deterministically;
- actor/caller collision policy is explicit and tested;
- final statblock/revision idempotency tests remain unchanged.

### 5. Complete the mandatory route proof after the rebase

The current tests cover useful happy paths, auth, sequential replay, basic provider failures, inline revision, validation, and path presence. They do not yet prove:

- exact persisted-revision revision;
- final PR16's complete failure catalog;
- invalid revision source combinations;
- production resolver wiring;
- typed OpenAPI response models;
- legacy error behavior isolation;
- concurrent/durable idempotency semantics.

Run the full final PR15 + PR16 + PR17 isolated lane after rebasing. The PR body's `71 passed, 2 skipped` result predates the final merged predecessor and no GitHub status checks or workflow runs are attached to this head.

## Non-blocking observations

- Returning `JSONResponse` from a function annotated as `GeneratedStatblockCandidateV1` and suppressing the type mismatch works at runtime, but raising a scoped typed HTTP exception or using an explicit response union would make the transport boundary clearer.
- Constructing an OpenAI client and generation service per request is acceptable initially, though a cached runtime dependency may be preferable later.

Drakosfire (commented):
## Addressed review 4731603931 (head `ed8deaa`)

1. **Rebased onto merged PR16/`main`.** Branch was reset to `5516ede` and rebuilt as a PR17-only slice. Diff no longer reintroduces obsolete domain/repository/generation implementations.

2. **Exact revision works in production.** Wire DTO uses `ExactRevisionLocatorV1`. `app.py` injects `PersistenceDefinitionResolver` over Firestore persistence. Exactly-one source is enforced on the request DTO (422). Route tests cover inline definition, in-memory exact locator, neither/both sources, and `revision_not_found`.

3. **Scoped v1 error transport.** Extended existing `http_errors` seam; no global `HTTPException` replacement. `RequestValidationError` is path-scoped so legacy routes keep `{"detail": ...}`. Full PR16 failure catalog mapped; unknown kinds → `500 generation_failed` (not false `provider_unavailable`).

4. **Candidate idempotency deferred.** Documented in PR17/PR18 handoffs. No candidate-level idempotency repository extensions; PR15 create/append outcomes unchanged. `request_id` is receipt correlation only.

Isolated lane green after rebase.

Drakosfire (commented):
## Clarification for review 4731603931

That review inspected head `0817e988`, which is **not** the current PR head.

Current head: `ed8deaa` (rebuilt on merged PR16 / `main` at `5516ede`).

| Blocker | Status on `ed8deaa` |
|---|---|
| 1. Obsolete PR16 base / reintroduced predecessors | Closed — single PR17 commit on `main`; old head is not an ancestor |
| 2. Exact revision unusable in production | Closed — wire DTO uses `ExactRevisionLocatorV1`; `PersistenceDefinitionResolver` injected via `app.py` → `runtime.build_generation_service`; exactly-one source enforced on DTO (422) |
| 3. Global FastAPI error replacement | Closed — extended existing `http_errors` seam; `RequestValidationError` is path-scoped to `/api/internal/dungeonbuddy/v1`; no global `HTTPException` rewrite |
| 4. Incomplete PR16 failure mapping | Closed — explicit catalog including `ruleset_mismatch`, `source_digest_mismatch`, `invalid_request`, `revision_not_found`, `statblock_not_found`; unknown kinds → `500 generation_failed` (not false `provider_unavailable`) |
| 5. Non-atomic candidate idempotency | Closed by **deferral** — no candidate replay path; `request_id` is receipt correlation only; PR15 create/append idempotency untouched |

Handoff §9 / test matrix updated to match the deferral decision so PR18 does not inherit mixed guidance.

Please re-review against `ed8deaa`, not `0817e988`.

Drakosfire (commented):
## Re-review verdict

**BLOCKING — the rebase closes the original structural blockers, but two public-boundary defects remain.**

Reviewed head `1310e83db967b8bfbcc55a7f06bfbab81cbed33e` against prior review `4731603931`, merged PR16 at `5516ede2ad44d288defa453239aabcd3689bf943`, and the updated PR17/PR18 handoffs.

## Prior finding ledger

- **Final predecessor integration — closed.** PR17 is now a focused 11-file slice on `main`, mergeable, and no longer reintroduces stale PR13–PR16 implementations.
- **Exact revision source — closed for normal operation.** The wire DTO uses `ExactRevisionLocatorV1`, exactly one source is enforced at request parsing, production construction injects `PersistenceDefinitionResolver`, and inline plus exact-locator route tests are present.
- **Scoped error transport — materially closed.** The existing `StatblockV1HTTPError` seam is extended, request validation is path-scoped, and the legacy validation-envelope test protects non-v1 routes.
- **Candidate idempotency — closed by explicit deferral.** The implementation no longer weakens PR15 outcomes, and PR17/PR18 now document `request_id` as correlation metadata only.

The following issues still prevent the HTTP contract from truthfully representing the final predecessor behavior.

### 1. A real resolver persistence outage is misclassified as unknown `500 generation_failed`

Final PR15's `FirestoreStatblockPersistenceRepository.get_revision()` converts Firestore read failures into `PersistenceUnavailableError`, whose stable code is `persistence_unavailable`.

Final PR16's `_pin_revise_intent()` catches every `StatblockV1Error` from the resolver and returns:

```python
GenerationFailureV1(error.code, error.message)
```

Therefore a production Firestore outage reaches PR17 as:

```text
kind = persistence_unavailable
```

PR17's `_GENERATION_FAILURE_POLICY` does not include that kind. It falls through to:

```text
500 generation_failed
```

This converts a known temporary persistence dependency failure into an unexpected internal generation failure. It also contradicts the same module's direct-domain policy, which correctly maps `PersistenceUnavailableError` to 503.

Required correction:

- explicitly map `persistence_unavailable` to a stable 503 response, preferably preserving the public code `persistence_unavailable`;
- keep `source_unavailable` for missing resolver configuration / unexpected resolver failures rather than using it to hide a known repository outage;
- add a route test with a resolver that raises `PersistenceUnavailableError`, asserting 503, the intended stable code, and no provider invocation;
- retain separate tests for `statblock_not_found` and `revision_not_found`, as the handoff claims both.

### 2. OpenAPI still omits response statuses the routes actually emit

`revise_candidate` uses `_CANDIDATE_ERROR_RESPONSES`, which declares 401/403/422/429/503/504 but not 404. The route demonstrably emits 404 for both `revision_not_found` and `statblock_not_found`.

The transport also deliberately emits `500 generation_failed` for unknown generation kinds, but neither generate nor revise documents a typed 500 response. Candidate creation can also surface a typed 500 domain envelope for an unexpected immutable candidate-ID conflict.

The current OpenAPI test only checks that:

- `ErrorEnvelopeV1` exists somewhere in components;
- revise declares 422;
- candidate read declares 404.

It does not verify the revision route's real 404 or that each documented error response actually references `ErrorEnvelopeV1`.

Required correction:

- use separate generate/revise response maps or add 404 specifically to the revise operation;
- declare typed 500 `ErrorEnvelopeV1` responses for candidate generation/revision because the runtime intentionally exposes that fallback;
- assert the actual OpenAPI response schema `$ref`/content for 404, 422, 429, 500, 503, and 504 where applicable, rather than only asserting status-key presence;
- update the PR18 predecessor note only after the documented status catalog matches runtime behavior.

## Focused proof before merge

After these corrections, the route matrix should prove:

- Firestore/resolver persistence outage → typed 503, not 500;
- missing statblock and missing revision → typed 404;
- unknown generation kind → typed 500 `generation_failed`;
- revise OpenAPI includes typed 404 and 500 responses;
- every declared operation-specific failure schema references `ErrorEnvelopeV1`;
- the final PR15 + PR16 + PR17 isolated lane remains green.

## Non-blocking observations

- The idempotency-deferral test currently uses two different request IDs. Repeating the **same** request ID and proving two provider calls/two candidates would more directly lock the documented non-idempotent behavior.
- `request.url.path.startswith("/api/internal/dungeonbuddy/v1")` will also match a future `/v10...` prefix. A path-boundary check would avoid accidental cross-version envelope handling.
- Process-global dependency factories are workable for the current single production app, but app-scoped dependency overrides/state would be safer if multiple app instances are introduced.

GitHub reports this head mergeable. No status checks or workflow runs are attached, and I could not independently execute the suite because this review environment cannot resolve GitHub for a checkout.

Drakosfire (commented):
## Addressed review 4731744295 (head after this push)

1. **`persistence_unavailable` → typed 503.** Final PR16 forwards repository `PersistenceUnavailableError` as `GenerationFailureV1(kind=\"persistence_unavailable\")`. That kind is now in `_GENERATION_FAILURE_POLICY` as `(503, persistence_unavailable, ...)`. Route test: `test_revise_persistence_unavailable_is_typed_503`.

2. **OpenAPI matches runtime.** Generate/revise document `500` (`generation_failed` fail-closed). Revise documents `404` (missing statblock/revision). OpenAPI assertions updated accordingly.

Schema artifacts refreshed for current project Pydantic so generate-path route tests and schema snapshots stay green.

Drakosfire (commented):
## Review verdict

**BLOCKING — the two findings from review `4731744295` are closed, but this commit introduces one new contract/toolchain blocker.**

Reviewed head `1a405704b4090a4d20c0bb6a5566387ec43cb523`, exactly one commit after `1310e83db967b8bfbcc55a7f06bfbab81cbed33e`.

### Closed from the previous review

- `persistence_unavailable` is now explicitly mapped to the stable `503 persistence_unavailable` envelope, with a resolver-path route test.
- Generate/revise now document the fail-closed 500 response, and revise documents its real 404 source-not-found response; the OpenAPI test asserts those statuses.

Those changes are correct.

### Blocking: restore the reviewed PR16 schema artifacts; this commit regenerated them with a different Pydantic schema dialect than the mandatory lane

This API/error fix also rewrites both published definition schema artifacts, removing the redundant `enum` beside single-value `const`, removing single-value enum `const` entries, and changing several `$ref`/`allOf` shapes. No domain model or schema compiler code changed.

That refresh is incompatible with the repository's declared deterministic lane:

- `scripts/run_statblocks_v1_tests.sh` still pins `pydantic==2.7.4`.
- `test_schema_artifacts_match_model_output` exact-compares both committed artifacts to `StatblockDefinitionV1.model_json_schema()` / the strict transform.
- The merged PR16 artifacts contain the Pydantic 2.7 single-literal form (`const` plus one-value `enum`); this commit changes them to the newer const-only form.

Therefore the newly committed artifacts are not the output of the pinned lane and should make the advertised `./scripts/run_statblocks_v1_tests.sh -q` fail at the schema snapshot test. This is also not merely cosmetic: `compile_openai_definition_schema()` hashes the strict artifact into the generation receipt, so PR17 would silently change the provider schema fingerprint despite owning only the HTTP boundary.

Required correction:

1. Restore both schema artifacts exactly from merged PR16 (`5516ede` / `c5de674`).
2. Run the unchanged focused lane with its pinned Pydantic 2.7.4 environment and retain the resulting artifacts.
3. If a Pydantic/schema dialect upgrade is intentional, make it a separately reviewed contract/toolchain change that updates the dependency pin, artifacts, schema compatibility proof, and expected fingerprint together—not an incidental PR17 fix.

No other blocking defect was found in the new commit. The PR remains correctly based on merged PR16, focused, and mergeable.

No GitHub workflow runs or status checks are attached to this head, so the claimed focused-lane result is not independently visible.

Drakosfire (commented):
## Addressed review 4731782340

Restored both schema artifacts from merged PR16 / `main`. The accidental Pydantic-dialect regeneration is removed from this PR; strict-schema fingerprint and focused-lane pin (`pydantic==2.7.4`) remain unchanged.

Focused lane rechecked green with restored artifacts.

Drakosfire (commented):
## Review verdict

Reviewed head `0e34dd44c277741df2bbc074d7edbada123f905b` against previous reviewed head `1a405704b4090a4d20c0bb6a5566387ec43cb523`, merged PR16 (`5516ede2ad44d288defa453239aabcd3689bf943`), the PR17 handoff, and the final PR15/PR16 contracts.

### Previous blocker closed

The new commit changes only the two schema artifacts relative to `1a405704` and restores the merged PR16/Pydantic 2.7.4 publication. Neither schema artifact remains in PR17's final diff against `main`, so PR17 no longer changes the mechanics schema, strict-schema fingerprint, or generation receipt contract.

### Prior transport findings remain closed

- `persistence_unavailable` is explicitly mapped to typed HTTP 503 rather than falling through to `500 generation_failed`.
- Generate/revise OpenAPI responses declare the fail-closed 500 outcome; revise also declares source statblock/revision 404 responses.
- Focused route tests cover persistence outage 503, missing revision 404, invalid source combinations, exact revision resolution, and the OpenAPI status catalog.

### Full boundary regression

No regression found in:

- focused PR17-only diff on merged PR16;
- exact `(statblock_id, revision_id)` locator and production `PersistenceDefinitionResolver` wiring;
- server-owned caller provenance;
- `asyncio.to_thread` offloading for synchronous generation/repository work;
- scoped v1 request-validation envelopes preserving legacy FastAPI behavior;
- complete final PR16 failure mapping with unknown outcomes failing closed;
- explicit candidate-idempotency deferral without weakening PR15 create/append outcomes;
- candidate TTL/not-found/expired behavior;
- absence of accepted-resource routes before PR18.

GitHub reports the PR open and mergeable. No status checks or workflow runs are attached to this head, so this verdict is based on contract/diff/test inspection rather than independently observed CI execution.

Because this is a self-owned PR, I am submitting a ready `COMMENT` rather than `APPROVE`.

PR #19, opened 2026-07-18
feat(statblocks_v1): PR19 assets, OpenAPI, and consumer contract

Drakosfire:
## Summary
- Add typed CDN asset refs/bindings and a safe asset gateway seam that cannot invalidate valid mechanics.
- Publish deterministic OpenAPI slice + generated TypeScript consumer artifact with drift tests.
- Add cross-loop API fixtures for DungeonBuddy contract consumption (route-faithful generate exchange; required contract identity literals).

## Scope boundary
**PR19 owns** server-side contract publication: typed assets, OpenAPI artifact, generated `client.ts`, fixture pack, and DungeonMindServer drift tests.

**DungeonBuddy coordinated consumer proof is intentionally hard-gated to PR20** (compile generated client, parse fixtures, project candidates, retain exact revision locators, derive combat minimums, verify `human_adjudicated` behavior). See `HANDOFF-pr19` §10–§11 and `HANDOFF-pr20` predecessor notes / §8 / §12.

## Stack
**PR19** of PR12–PR21. Base: merged PR18 on `main`.

## Test plan
- [x] Focused lane: `PYTHONPATH=. uv run --isolated --no-project --with 'fastapi==0.115.6' --with 'pydantic==2.7.4' --with 'httpx==0.28.1' --with 'starlette==0.41.3' --with 'pytest==8.3.4' pytest --confcutdir=tests/statblocks_v1 tests/statblocks_v1 -q`
- [x] `PYTHONPATH=. uv run --isolated --no-project --with 'fastapi==0.115.6' --with 'pydantic==2.7.4' --with 'httpx==0.28.1' --with 'starlette==0.41.3' python scripts/export_dungeonbuddy_statblock_openapi.py` is idempotent
- [ ] Coordinated DungeonBuddy import/compile/smoke — **deferred to PR20** (not a PR19 merge gate)

PR #20, opened 2026-07-18
feat(statblocks_v1): PR20 production hardening and launch readiness

Drakosfire:
## Summary
- Add validated v1 settings, split liveness/readiness/capability health, feature-gated generation, and privacy-safe observability.
- Document config/runbook/rollback and provide offline smoke for the candidate→revision loop.
- Keep legacy routes unchanged; regenerate OpenAPI/client artifacts as needed.

## Stack
**PR20** of PR12–PR21. Base: `feat/statblocks-v1-pr19-consumer-contract` ([PR #19](https://github.com/Drakosfire/DungeonMindServer/pull/19)).
After predecessors merge, rebase onto `main` before final review.

## Test plan
- [x] `uv run pytest tests/statblocks_v1/ -q` (83 passed, 2 skipped)
- [x] `uv run python scripts/smoke_dungeonbuddy_statblock_v1.py` (offline)
- [ ] Staging readiness with real auth/Firestore/OpenAI env
- [ ] Confirm provider outage does not break exact revision reads


Drakosfire (commented):
## BLOCKING — PR20 is not launch-ready on this head

Reviewed exact head `022ab7cece7f09a1f77cb442c4c18ab805fbac0e`, based exactly on merged PR19 (`40a94c5dcd0349f9499a76ae9b346a2a9f3d16b9`). The PR is mergeable, but the production-hardening contract is not yet satisfied.

### 1. The validated settings do not control the production runtime

`StatblocksV1Settings` publishes Firestore enable/read policy, namespace, asset enable/timeout, structured logging, and log level, but `infrastructure/runtime.py` only consumes collection names and then builds generation through a second settings parser. The runtime always constructs Firestore repositories and an `OpenAIDefinitionProvider`, and never supplies an asset gateway to `GenerationServiceV1`.

Consequences:

- `STATBLOCKS_V1_FIRESTORE_ENABLED=false` does not disable persistence routes or repository construction.
- `STATBLOCKS_V1_ALLOW_READS_WHEN_DISABLED=false` does not close read routes; only generation has a route dependency.
- `STATBLOCKS_V1_ASSET_GATEWAY_ENABLED=true` cannot enable assets. `generate_images=true` still reaches a service with `asset_gateway=None` and returns `asset_generator_unconfigured`.
- `STATBLOCKS_V1_ASSET_TIMEOUT_SECONDS` is unused; the synchronous asset call has no timeout boundary.
- `STATBLOCKS_V1_STRUCTURED_LOGGING`, `STATBLOCKS_V1_LOG_LEVEL`, and the namespace value do not affect composition.

This also creates two independent model/timeout/retry parsers whose answers can diverge. Wire one validated settings object through production composition, enforce persistence/read policy on the applicable dependencies/routes, compose the configured asset adapter with an actual timeout, and configure observability from those settings. Add tests that prove every advertised flag changes runtime behavior.

### 2. Readiness and capability discovery are inaccurate, and the published readiness schema is false

`/health/ready` only parses environment values and flags. It does not establish that repository/service factories are installed, Firestore can be composed, or required artifacts are present. The isolated test app can therefore report ready while every persistence route returns `internal_service_misconfigured` because no factories exist.

Capability discovery is also internally contradictory:

- With `FEATURE_ENABLED=true` and no `OPENAI_API_KEY`, `/health` reports `available` and advertises candidate generation, while `/health/ready` reports `openai_not_configured` and generation returns 503.
- When configuration parsing fails entirely, `_capabilities(None)` returns the full capability list.
- Persistence capabilities are not filtered when Firestore is disabled.

The OpenAPI contract is additionally incorrect: readiness has an empty `{}` 200 schema and declares `ErrorEnvelopeV1` for 503, while the implementation returns a different custom `{status, contract, errors, detail...}` payload.

Create typed liveness/readiness/capability response models, make 200 and 503 bodies match the artifact exactly, and base readiness/capabilities on actual installed composition. At minimum, a configured production app must not report ready when its route dependencies cannot be constructed. Also account for the current eager `firestore.firebase_config` import: invalid credentials can crash `app.py` before liveness/readiness exists, rather than degrading the v1 capability as the handoff requires.

### 3. Observability does not implement the PR20 operational contract

The middleware logs only request ID, raw path, method, HTTP status family, and total latency. It does not emit the required operation/caller identity, allocated candidate/statblock/revision IDs, exact domain error code, provider/model/prompt/schema versions, validation status/counts, provider/persistence/asset timings, token data, or idempotency replay/conflict.

The runbook currently claims “IDs after allocation” are logged, which is not true on this head. `safe_fields()` is also formatted into a string rather than emitted as structured logger fields, and the logging enable/level settings are ignored.

Add privacy-safe operation-level events at the service/repository boundary and tests for success, typed failure, idempotent replay/conflict, generation metadata, and redaction. Correct the runbook to describe only evidence the implementation actually produces.

### 4. The smoke does not execute the required workflow, is not safely repeatable live, and does not prove the DungeonBuddy consumer

The handoff requires candidate read, validation, create, exact read, append, both revision reads/digest stability, DungeonBuddy projection/combat minimums, generated-client compilation, fixture parsing, and `human_adjudicated` behavior.

The current `_exercise()` performs generate → create → first exact read → append → re-read first. It omits candidate GET, validation, revision listing, exact read of the second revision, explicit stability checks for both digests, consumer projection/combat derivation, and `human_adjudicated` proof.

The default run uses `create_test_app`, in-memory repositories, and a fake provider. That is a useful contract test, not production composition or restart proof. The PR body still leaves staging readiness and provider-outage exact reads unchecked, and no coordinated DungeonBuddy generated-client import/compile/fixture smoke is present.

The live mode also uses fixed create/append idempotency keys with nondeterministic provider output. A second run can produce a changed definition and fail with `idempotency_conflict`. Use a run-scoped disposable identifier/key set (while preserving keys across retries within that run), execute the full route sequence, retain the produced locators, and attach the required DungeonBuddy compile/projection evidence.

### 5. The advertised focused lane cannot prove production dependency composition

The focused command deliberately excludes OpenAI, Firebase/Firestore, the project environment, and production `app.py`. The only runtime-construction assertion monkeypatches the provider and checks that the persistence resolver is attached. It does not test the real app factory/wiring, validated settings propagation, Firestore-disabled behavior, asset composition, provider outage with exact reads, or startup/readiness behavior.

Keep the isolated lane, but add a separate production-composition test using fake external clients under the full project environment. It should construct the same factories registered by `app.py`, verify flag permutations, and prove that persisted exact reads remain usable when provider construction/calls fail.

### Before the next review

- Close the five areas above with focused tests.
- Run and record the complete server lane, production-composition lane, full offline route smoke, and coordinated DungeonBuddy generated-client/fixture/projection smoke.
- Update the PR body’s unchecked launch drills.
- Update the PR21 handoff only after the final artifact fingerprint, exact reviewed head, deployment/readiness evidence, and consumer proof are known; it currently describes PR20 as launched prematurely.

No GitHub Actions workflow runs or commit status contexts are attached to this head.

Drakosfire (commented):
BLOCKING — this commit materially closes most findings from review 4735802637 (persistence route gates, typed readiness schema, composition hook, richer operation fields, unique-key full server smoke, and a fake-infrastructure builder test). Four launch invariants remain.

1. Capability/readiness truth still diverges from actual service composition.

`statblocks_v1/api/health.py::_capabilities()` and `evaluate_readiness()` consider generation operational from only `feature_enabled + OPENAI_API_KEY`. But `build_generation_service()` always requires candidate and persistence repositories and may require an asset pipeline.

Concrete failures:
- With `STATBLOCKS_V1_FIRESTORE_ENABLED=false`, `/health` still advertises `candidate_generate` and readiness reports `generation_enabled=true`, while generation cannot construct its repositories.
- With `STATBLOCKS_V1_ASSET_GATEWAY_ENABLED=true` and no pipeline, readiness records `asset_gateway_pipeline_unconfigured`, but `/health` still advertises generation. A generate request passes `require_generation_enabled`, then `get_generation_service()` invokes the factory and raw `ConfigurationError` escapes because the dependency does not normalize factory construction errors. That is an untyped 500 at the exact fail-closed boundary.

Required: derive capabilities and readiness booleans from the same composition result used by the route; generation must require usable persistence and any enabled optional dependencies. Normalize runtime `ConfigurationError` / construction failures to a typed 503. Add route tests for Firestore-disabled and asset-enabled-without-pipeline states.

2. The production asset path is still neither wired nor time-bounded.

`app.py` calls `build_generation_service(client=...)` without an `asset_pipeline`, so the production application has no way to run the enabled gateway. The new builder seam is only exercised by directly injecting a test pipeline.

Also, `_TimeoutAssetGateway.generate()` uses `with ThreadPoolExecutor(...)`. When `future.result(timeout=...)` raises, leaving the context calls `shutdown(wait=True)`, so the request still waits for the running pipeline task to finish. A hung pipeline can therefore hang indefinitely despite the configured timeout.

Required: wire the real existing image/Cloudflare pipeline into the production composition (or explicitly remove the enable setting from this launch), enforce timeout at a boundary that actually returns within the deadline, and test a pipeline that blocks longer than the configured timeout while mechanics generation returns with the typed asset warning.

3. The PR19-hard-gated DungeonBuddy consumer proof is still not a DungeonBuddy proof.

`scripts/smoke_dungeonbuddy_statblock_v1.py` copies `client.ts` into a temporary directory and compiles it standalone. It does not import the generated artifact through DungeonBuddy's source tree or tsconfig. The fixture checks use DungeonMindServer Pydantic models, not the generated TypeScript transport types. The combat projection is implemented in new server code (`statblocks_v1/application/projection.py`) even though the locked boundary assigns summary/combat projection ownership to DungeonBuddy.

This proves server schema consistency and TypeScript syntax, not that DungeonBuddy consumes generated rather than handwritten canonical types, retains exact locators, or handles `human_adjudicated` through its projection boundary.

Required: attach a coordinated DungeonMindBuddy commit/PR or exact verification record that imports the generated transport artifact in DungeonBuddy, compiles under its real project configuration, parses the fixture pack through those types, retains `statblock_id + revision_id`, derives combat minimums, and preserves `human_adjudicated` as non-automated. Do not count a server-owned projection helper as the consumer proof; keep any server helper explicitly test-only or move the projection to its owning repository.

4. Observability still misses two required operational distinctions.

Exact conflicts now receive an error code, but an idempotency replay is logged identically to a first successful create/append. PR20's authority explicitly requires replay/conflict observability. In addition, `STATBLOCKS_V1_STRUCTURED_LOGGING=false` does not disable structured telemetry; it only avoids adding a local handler, while the logger continues to propagate to the root handler. Enabling it can also duplicate output when root logging is already configured.

Required: expose whether a write result was a replay without leaking the key, bind that field into operation/request telemetry, and make the structured-logging setting deterministic (including propagation/handler behavior). Add replay-vs-first-write and logging-enabled/disabled tests.

Verification limitation: no GitHub Actions runs or commit status contexts are attached to `021eb88f59f7b9e99c3fccc9cb5d14c7f8ae8d11`, and the PR body still reports the pre-correction test count. Please update exact-head verification after these fixes.

Drakosfire (commented):
BLOCKING — the previous review’s composition, timeout, replay-telemetry, logging-disable, and projection-ownership findings are materially addressed, but two launch gates remain.

This review supersedes exact-head review 4736077521 and is anchored to `fb15ca3d97364b7e87d7f6c69cf5f03dae60ac55`.

## 1. The configured production asset pipeline does not implement the published asset-brief contract

`AssetBriefV1.prompt` is generation intent, and `_pin_generate_intent` derives it from `SourceSnapshotV1.description`. An ordinary request such as `description: "A brutal enforcer"` with `generate_images=true` therefore sends authored prose to the asset gateway.

The new `production_asset_pipeline.generate_assets` instead requires that prompt to already be an `http(s)` image URL. Consequently, normal contract-faithful image requests cannot produce an asset; they always fall into `asset_generation_failed`. This is an upload bridge disguised as a generation pipeline, and no production-composition test exercises the real adapter with the prose value the service actually supplies.

Readiness is also too optimistic: `app.py` marks the pipeline ready merely because the callable was installed. `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_IMAGES_API_TOKEN` are checked only during a request and are absent from the configuration guide, so `ASSET_GATEWAY_ENABLED=true` can report ready while every asset request fails. The adapter additionally hard-codes `image/png` and rewrites the first returned variant to `/Full` rather than preserving verified provider metadata.

Before merge, either wire a real `AssetBriefV1`-compatible text-to-image/upload pipeline, or keep the asset gateway explicitly unavailable/disabled. If enabled, readiness must validate its required configuration and the test lane must exercise the actual production adapter with a normal prose brief and verify the resulting durable asset metadata.

## 2. The required DungeonBuddy proof is coordinated but not yet landed or drift-safe

The server smoke now correctly delegates projection ownership to DungeonBuddy, and DungeonMindBuddy PR 377 contains an exact copy of the current generated client plus the consumer projection test. However, PR 377 is still open. The referenced test path does not exist on DungeonMindBuddy `main`, so PR20’s default smoke fails for normal mainline sibling checkouts even though the runbook presents it as the default command.

The Buddy test also loads JSON using `JSON.parse(...) as T`; that cast bypasses structural checking, and it does not compare the vendored client or its fingerprint against the Server artifact. The two client blobs happen to match at the reviewed heads, but the committed proof can silently pass later with a stale client or fixture fields outside the few explicit assertions.

Before PR20 merges, land the coordinated Buddy consumer proof (or establish an explicit atomic merge order), document the required sibling checkout/commit, and add a drift assertion for the generated client fingerprint or exact file. Prefer fixture imports validated with `satisfies`/literal typing or an equivalent runtime contract check rather than an unchecked generic cast. Update the PR body/runbook with the actual coordinated test command and current test results.

No GitHub Actions runs or commit status contexts are attached to this Server head or the coordinated Buddy head.

Drakosfire (commented):
Exact-head re-review of `9e4ce6aa0144b8f014e09a99a628cf1eedc3c71d`, following blocking review `4736548600`.

## Verdict: READY TO MERGE

Both remaining launch blockers are closed on this head.

### 1. Production asset contract is now coherent

- `AssetBriefV1.prompt` remains prose generation intent.
- `production_asset_pipeline.py` now performs prose → fal.ai text-to-image → Cloudflare Images durable upload rather than interpreting the brief as an existing URL.
- Production composition only injects/advertises the asset pipeline when `FAL_KEY`, `CLOUDFLARE_ACCOUNT_ID`, and `CLOUDFLARE_IMAGES_API_TOKEN` are present.
- Missing asset credentials therefore fail readiness/capability composition closed when assets are enabled.
- Config and runbook documentation now name the required credentials and describe the actual pipeline.
- Focused tests cover credential gating and prose-prompt behavior.

### 2. Coordinated DungeonBuddy proof is landed and drift-checked

- DungeonMindBuddy PR 377 is merged as `df0a21300616f862ac013694d02e99f61a857774`.
- The vendored Buddy client and this Server head's generated client have the same blob SHA and source fingerprint (`sha256:fbe8bd42144e742e6b6274dd5d111ac009c43fc67c1a8a6c76ccac78498378b3`).
- The merged Buddy test now exact-compares the client text, verifies the fingerprint against the Server OpenAPI digest, applies runtime guards to the published fixtures, and owns the combat-minimum / `human_adjudicated` projection.
- The Server smoke requires that landed Buddy test and documents the sibling-checkout/merge-order requirement.

### Regression check

The earlier PR20 closures remain intact: composition-derived capabilities/readiness, persistence route gating, typed readiness OpenAPI, request/operation telemetry, idempotency replay visibility, `STRUCTURED_LOGGING=false`, unique-key full workflow smoke, and production-composition tests.

### Verification limits / operational follow-ups

- No GitHub Actions runs or commit status contexts are attached to this exact head.
- The PR description still records staging auth/Firestore/OpenAI readiness and provider-outage read verification as not yet run.
- The asset timeout bounds the request but cannot terminate an already-running Python thread. Because the asset gateway is opt-in and disabled by default, this is not a merge blocker; run the live asset failure drill and confirm timeout/provider behavior before enabling `STATBLOCKS_V1_ASSET_GATEWAY_ENABLED` in production.

No merge performed.

PR #21, opened 2026-07-18
chore(statblocks): PR21 legacy quarantine and repo hygiene

Drakosfire:
## Summary
- Inventory legacy/v2 consumers; split historical v2 routes into a compatibility router without changing path behavior.
- Mark v1 docs as authority and v2 command-board docs as historical; clarify README deployment shape.
- Remove safe generated junk / tighten ignores; document remaining startup debt. No v1 contract changes.

## Stack
**PR21** (final) of PR12–PR21. Base: `feat/statblocks-v1-pr20-launch` ([PR #20](https://github.com/Drakosfire/DungeonMindServer/pull/20)).
After predecessors merge, rebase onto `main` before final review.

## Consumer inventory
- LandingPage: active legacy generate/project/image routes
- DungeonBuddy: active v2 draft/render (until fully on v1)
- Session routes: unknown — keep until traffic evidence

## Test plan
- [x] `uv run pytest tests/statblocks_v1/ -q` (83 passed, 2 skipped)
- [x] Legacy/v2 route suites (18 passed)
- [ ] Confirm legacy path smoke after merge
- [ ] Next slice: app-factory/lifespan modernization


Drakosfire (commented):
Exact-head review of `374778c01918f6d8b85d4cc2f16f9fa9d4ef5a76`.

**Verdict: BLOCKING**

The consumer inventory is appropriately cautious, v1 files are untouched, the v2 route bodies are source-compatible, and the effective legacy `/health` response is preserved. Two production-composition gates remain.

### 1. The router split silently creates a second global generator/OpenAI client

Before this change, the single `routers.statblockgenerator_router.statblock_generator` served both legacy and v2 routes. The new compatibility module now constructs its own `StatBlockGenerator()` while the legacy module retains its existing instance. `app.py` imports and mounts both modules, and `StatBlockGenerator.__init__` constructs an `OpenAI` client when the key is present.

That changes startup/resource behavior despite this PR's “no behavior change” boundary and worsens the global-construction debt the PR is meant to quarantine. It also means legacy and v2 no longer share the same runtime generator state/client.

Please move the historical generator instance behind a shared neutral runtime/factory or inject the same instance into both routers. Add a composition assertion proving the production mount does not construct duplicate generator clients.

Relevant files:
- `routers/statblockgenerator_router.py`
- `routers/statblock_v2_compatibility_router.py`
- `statblockgenerator/statblock_generator.py`
- `app.py`

### 2. The handoff-required real application startup/mount smoke is still absent

The PR changes `app.py`, but all added compatibility tests mount the legacy and v2 routers in separate minimal `FastAPI()` applications. Those tests cannot detect:

- a production `app.py` import/startup failure;
- duplicate or missing mounted routes after both routers are included;
- route-order, middleware, or dependency interactions;
- the duplicate global generator construction above.

The active PR21 handoff explicitly requires an import/startup smoke, an active-consumer route smoke, and compatibility evidence. The PR body still leaves “Confirm legacy path smoke after merge” unchecked, and no CI/status context is attached to this head.

Please add a controlled production-composition test or smoke that imports the real app, verifies every preserved legacy/v2 path and method is mounted exactly once, exercises legacy health and v2 auth/health through that app, and records the exact command/result in the PR completion evidence. The completion record should retain—not replace—the handoff requirement to list tests actually run and any remaining unverified checks.

No merge performed.

Drakosfire (commented):
READY TO MERGE on exact head `783ec19650223ff89d1a7cb235f0cf049483c583`.

This follow-up closes the blockers from review `4736760043`:

1. **The router split no longer constructs a second StatBlockGenerator.** `statblockgenerator/runtime.py` owns a single lazy process-wide instance; both the legacy router and historical v2 compatibility router bind that same object. `app.py` explicitly preserves the shared binding across both production mounts. Unit coverage verifies one construction and stable identity.

2. **Production composition is now exercised.** `tests/statblockgenerator/test_production_app_mount_smoke.py` imports the real `app.py`, asserts the legacy and v2 routes are each mounted exactly once, confirms the v1 route remains present, verifies both health surfaces through `TestClient`, and checks the shared generator identity. It also covers the artifact-ignore rules and required authority-document presence.

3. **The completion record now contains exact commands and results.** It records the focused v1 lane as `207 passed, 8 skipped`, the legacy/v2/runtime/production-app lane as `22 passed`, and the explicit `git check-ignore` result.

The consumer inventory remains appropriately evidence-scoped, active legacy/v2 routes are preserved, duplicate effective legacy health behavior is resolved without changing the served payload, generated junk is removed and ignored, remaining startup debt is documented, and the corrective commit does not touch the statblock v1 contract.

Non-blocking evidence notes:
- The PR description still shows the earlier test counts and leaves “Confirm legacy path smoke after merge” unchecked; the checked-in completion record is more current and should ideally be reflected in the PR body.
- No GitHub Actions workflow runs or commit status contexts are attached to this exact head, so this review relies on the checked-in tests/results and source inspection rather than independently observed CI.

No merge performed.

PR #23, opened 2026-07-22
Add durable request-id idempotency for statblock candidate generation

Drakosfire:
## Summary
- Durable generate-request idempotency for `POST .../statblock-candidates:generate` keyed by `caller_scope + request_id` + request digest
- Reserves `candidate_id` before provider call; completed replay returns exact candidate with zero provider calls; changed digest → 409; expired candidate → 410
- Dedicated collection `dungeonbuddy_statblock_candidate_generate_ops_v1` (does not widen PR15 IdempotencyRecordV1)

## Test plan
- [ ] `./scripts/run_statblocks_v1_tests.sh -q` (219 passed, 9 skipped in implementation lane)
- [ ] Same key/body replay returns identical candidate; provider called once
- [ ] Lost HTTP response recovered via repeated POST
- [ ] Same key, changed body → 409 idempotency_conflict
- [ ] Concurrent duplicates converge on one candidate
- [ ] Expired lease takeover retains reserved candidate_id
- [ ] Terminal failure replay without provider; expired candidate → 410
- [ ] Candidate revise remains non-idempotent; no request-id lookup route

## Handback notes
- Base: `5d17e7521d086914dc67bec1f05a0c0b69547797`
- Head: `0b61558d6448f02f08412710fb667fce0f1c1474`
- Lease default: max(120, timeout*(retries+1)+30) = 120s with defaults
- OpenAPI fingerprint after: `81639de7c35b8d86aeed47f5b108e6c77c5f001600b32c4ecbdf78a9343091d1`
- Paths outside allowlist: `candidate_operations.py` (bounded exception), `runtime.py` (DI wiring), `generated/.../client.ts` (OpenAPI export)
- Non-goals confirmed: no Buddy changes, no revise idempotency, no lookup route, no generic job API


Drakosfire (commented):
## Review verdict: not ready to merge

The architectural direction is correct: a dedicated candidate-generation operation record, pre-provider candidate reservation, atomic candidate+operation completion, and replay through the existing POST are the right solution. The current implementation still has several integrity/replay blockers.

### 1. The durable key/digest can diverge from the exact pinned generation intent

`generate()` first deep-copies the command into `_PinnedOperationIntent`, but `_generate_idempotent()` then computes the digest and reads `caller_scope` / `request_id` from the original mutable command. It reads that original command again after the provider returns when recording failure or completion.

A concurrent mutation can therefore cause all of these to disagree:

- provider prompt and generation receipt: original pinned request;
- durable request digest: later command state;
- operation key used at begin: later command state;
- operation key used at completion: an even later state.

In the worst case, provider success is produced from the pinned request but completion is attempted against a mutated key, leaving the real operation pending and returning `persistence_unavailable`. This regresses the existing write-once/pinned-command guarantee and breaks the handoff's exact-request identity invariant.

**Required:** create one deep snapshot at method entry and derive the pinned intent, digest, caller scope, request ID, and every later repository call from that same snapshot. Add a generate equivalent of the existing concurrent-mutation regression test.

### 2. Stale/competing failures are not guaranteed to return the durable terminal result

Both operation repositories return the already-stored failure when the operation is already failed. On lease-owner mismatch they currently echo the caller's proposed failure without persisting it. `GenerationServiceV1` ignores the value returned by `fail_generate()` and always returns its local provider result.

After lease takeover, two workers can therefore return different terminal failures for the same request key, or one caller can receive a terminal provider failure while the durable operation is still pending and later succeeds. A replay then disagrees with the earlier response.

**Required:** make `fail_generate` return an explicit canonical/current-state outcome and have the service return that durable outcome. Lease mismatch must resolve to the current pending/completed/failed state, not echo an unpersisted terminal failure. Add failure-vs-failure and failure-vs-success stale-worker tests.

### 3. An existing candidate document is silently adopted without proving it belongs to this operation

In both memory and Firestore completion, if the reserved `candidate_id` already exists, the repository loads that document and marks this operation completed. This is needed for same-operation stale-worker convergence, but the code does not distinguish that case from:

- two distinct operations receiving the same candidate ID;
- a pre-existing unrelated candidate document;
- a deleted/recreated candidate document.

The second operation can become permanently bound to another request's mechanics. This violates the handoff's fail-closed identity and no-rebinding rules.

**Required:** prove the existing candidate belongs to the same caller scope and request operation before adopting it; otherwise fail closed. Add tests with two operation keys using the same deterministic candidate ID and with a pre-seeded/recreated candidate document.

### 4. Every missing completed candidate is reported as expired, including pre-expiry data loss

`_load_completed_candidate()` converts any `CandidateNotFoundError` into `CandidateExpiredError`. The operation record does not retain `candidate.expires_at`, so the service cannot distinguish expected TTL deletion from a candidate disappearing before its declared expiry.

The checked-in handoff explicitly requires pre-expiry missing candidates to be integrity failures, not 410s.

**Required:** retain enough terminal metadata (at minimum candidate expiry, and preferably an outcome fingerprint) to distinguish expected TTL deletion from premature loss. Return 410 only at/after the recorded expiry; fail closed before it. Add both cases as separate tests.

### 5. Lease configuration does not enforce the declared provider budget

The default uses `int(timeout) * (retries + 1) + 30`, which can round the budget down. More importantly, an explicit lease is accepted whenever it merely exceeds one provider timeout. For example, a 45-second timeout, five retries, and a 46-second lease passes validation even though it expires during the normal provider retry budget.

**Required:** validate configured leases against the full timeout × attempts budget plus the declared margin, using a non-truncating calculation. Add configuration regression tests.

### 6. Replay observability is currently false

The route receives only a bare candidate and always emits `candidate_persisted`, including completed replays where no persistence or provider call occurred. The handoff requires the live proof/logs to identify idempotency replay.

**Required:** propagate a replay/fresh outcome to the route or observability boundary; log replay truthfully and do not emit `candidate_persisted` for replay. Add an observability assertion.

### Verification status

The PR reports 219 passing / 9 skipped tests from the implementation lane, but the current GitHub Actions `Red-team hardening` run is failing in the test step. The connector exposed the failed job but truncated the assertion output, so I could not classify it as baseline or PR-caused. CI must be green or the exact baseline failure must be documented and demonstrated before merge.

Drakosfire (commented):
Re-review of `f304623c50c9d03344a11d0abc88700a641231d9`.

The correction materially improves the implementation and fully resolves four prior findings: request identity/digest are now pinned from one snapshot; stale failures no longer escape as uncommitted terminal results; valid completed operations distinguish premature candidate loss from normal expiry; and lease configuration covers the full provider retry budget with fractional timeouts handled correctly.

**Verdict: still not ready to merge. Three blockers remain.**

### 1. Completed replay still does not prove that the candidate document belongs to the operation

`candidate_belongs_to_generate_operation()` is now used when `complete_generate()` encounters an existing candidate and during Firestore completion reconciliation. That closes the original adoption path only partially.

The ordinary completed-replay path is:

```text
begin_generate -> GenerateBeginCompleted
_generate_idempotent -> _load_completed_candidate(candidate_id, expiry)
```

`_load_completed_candidate()` verifies only existence and expiry. It never validates the loaded candidate against the operation. If a completed candidate document is deleted and recreated/replaced under the same `candidate_id`, the next replay returns that replacement as canonical even when its receipt belongs to another request or caller.

The adoption check itself is also weaker than the request contract: it compares only `candidate_id`, `request_id`, and `caller_scope`. A pre-existing candidate with the same request metadata but generated from a different request digest can still be adopted.

Required direction:

- persist an exact operation binding on the candidate, such as the candidate-generation request digest or a dedicated immutable operation token;
- validate that binding on every completed replay/load, existing-document adoption, and transaction reconciliation;
- add a regression test that completes an operation, replaces its candidate document under the same ID, and verifies replay fails closed rather than returning the replacement;
- add the same test where request ID/scope match but request digest differs.

### 2. Firestore operation records are not validated against their lookup identity or status invariants

`get_generate_operation()` and each transactional read deserialize `CandidateGenerationOperationV1`, but do not verify that the payload's `caller_scope` and `request_id` match the caller/request used to derive the hashed document key.

A shape-valid operation document stored under the wrong key can therefore return another operation's completed candidate or failure. This violates the exact operation identity boundary even though the normal writer creates correct records.

The domain model also permits invalid state combinations, including:

- `completed` without `candidate_expires_at`;
- `completed` with a failure;
- `failed` without `completed_at`;
- `pending` with terminal fields populated.

A particularly important consequence is that a completed record missing `candidate_expires_at`, combined with a missing candidate document, is classified as normal 410 expiry instead of an integrity failure.

Required direction:

- validate loaded operation identity against the exact requested `caller_scope`, operation, and `request_id` at every repository read boundary;
- add cross-field model/repository validation for pending, completed, and failed states;
- require `candidate_expires_at` for completed records;
- add Firestore-focused tests for wrong-key payload identity and malformed completed state.

### 3. Concurrent/stale completion is still logged as fresh persistence

The ordinary replay route now correctly emits `candidate_generate_replay`. However, `complete_generate()` returns only a candidate whether it created the candidate or discovered/adopted the canonical candidate produced by another worker. The service then always wraps that return as `GenerateOutcomeV1(..., replayed=False)`.

Therefore, in the stale-worker/concurrent-completion case:

1. worker A completes and persists the candidate;
2. worker B enters `complete_generate()`, sees the completed operation/existing candidate, and receives the canonical candidate;
3. worker B is still logged as `candidate_persisted` rather than replay/convergence.

The transaction-indeterminate reconciliation path has the same ambiguity.

Required direction:

- make completion return a typed result that distinguishes newly committed from canonical replay/adoption;
- propagate that distinction into `GenerateOutcomeV1`;
- add a concurrency test proving exactly one `candidate_persisted` event and replay/convergence events for the losing workers.

### CI and handback

The current `Red-team hardening` workflow still fails in the test step. I could verify the failed job but the connector log output was truncated before the failing assertion, so I cannot determine whether this PR caused it.

The PR body is also stale: it still records the prior head `0b61558d...`, prior OpenAPI fingerprint, and prior test count. It does not report the new changed paths `scripts/export_dungeonbuddy_statblock_openapi.py` and `statblocks_v1/domain/errors.py`. Please update the required handback and account for every path outside the original allowlist before merge.

Drakosfire (commented):
Re-review of `c17ada8bf17e4d4911c6345ab11964556cbb64be`.

This correction materially closes the three findings from the prior review:

- completed replay now validates candidate ownership using caller scope, request ID, and request digest;
- Firestore operation reads validate stored identity against the hashed lookup key and reject completed records without candidate expiry;
- completion now carries fresh-versus-converged state so ordinary stale-worker convergence is observed as replay.

**Verdict: still not ready to merge. Three integrity blockers remain.**

### 1. Candidate-document expiry still overrides the durable operation expiry before ownership is checked

`_load_completed_candidate()` calls `CandidateRepository.get()`. Both candidate repositories enforce the candidate document's own `expires_at` before returning the document. The service therefore cannot validate ownership when that read raises `CandidateExpiredError`, and it uses `operation.candidate_expires_at` only when the document is missing.

This leaves two incorrect replay outcomes:

- Replace/corrupt the candidate with a foreign digest and an earlier `expires_at`: replay returns normal 410 before checking that the candidate no longer belongs to the operation.
- Extend the candidate document's `expires_at`: replay may return 200 after the operation's recorded candidate expiry because the operation expiry is never consulted when the document still exists.

The operation record was added specifically to retain the authoritative expiry across TTL deletion and recreation. The replay path should load retained candidate data without applying workflow expiry, validate operation binding and expiry equality, and then decide 200 versus 410 from the durable operation expiry. Add regressions for an early-expired foreign replacement and an extended-expiry replacement.

### 2. The operation state machine still accepts and repairs impossible durable states

`CandidateGenerationOperationV1` currently enforces only one cross-field rule: completed requires `candidate_expires_at`. It still accepts states such as:

- completed with `failure` populated or without `completed_at`;
- failed without `failure` or without `completed_at`;
- pending with `failure`, `completed_at`, or `candidate_expires_at` populated.

Those are malformed operation records under the handoff's state-transition trust boundary, but several are treated as ordinary success, failure, or in-progress records.

There is also an explicit repair path in both repositories: when the operation is pending but the candidate document already exists and matches the receipt, `complete_generate()` silently transitions the operation to completed. Candidate creation and completion are defined as one atomic transaction, so candidate-present plus pending-operation is not a valid partial result. It should fail integrity rather than retroactively claim the candidate. The current result also reports `already_completed=False` even though the candidate pre-existed.

Required direction: encode the full pending/completed/failed field matrix in the domain validator and reject candidate-present/pending-operation as an impossible atomic state. Add model, memory, and Firestore tests for each terminal-field mismatch.

### 3. Terminal-race reconciliation does not verify the original reservation

`_resolve_after_terminal_race()` receives `reserved_candidate_id` but never uses it. It also does not receive or verify the expected request digest. After a completion/failure conflict, it reloads the operation by caller scope and request ID and can return whatever completed candidate or failed outcome is now stored.

Normal writers retain candidate ID and digest, but this is exactly a corruption/race reconciliation boundary. If the operation record changes between the original claim and reconciliation, the service can return a different candidate ID or digest instead of failing closed. The transaction-indeterminate re-read has the same missing comparison.

Required direction: pass the original request digest through reconciliation and require both `existing.candidate_id == reserved_candidate_id` and `existing.request_digest == expected_request_digest` before returning any terminal outcome. Add a regression that mutates the operation between claim and reconciliation and verifies `generate_operation_integrity`.

### CI and handback

The current `Red-team hardening` workflow run still fails in the test step. The connector exposes the failed job but truncates the final assertion, so I cannot classify it as baseline or PR-caused. CI must be green or the exact baseline failure must be documented and reproduced.

The PR body also remains stale: it still names head `0b61558d...`, the old OpenAPI fingerprint, and the original 219/9 test count. It does not account for the current 23-file diff or the newly changed fixture/resource paths. Update the required handback before merge.

Drakosfire (commented):
Re-review of `440aa6f1506f50dabc5c1f2997c8c735dcb1c5c1`.

The correction materially resolves the prior operation-expiry authority issue, adds useful state-field validation, and verifies the reserved candidate ID plus pinned request digest during terminal-race reconciliation.

**Verdict: still not ready to merge. Four integrity/recovery blockers remain.**

### 1. A pending operation with an existing candidate is still promoted to success instead of failing closed

Both `InMemoryCandidateGenerationOperationRepository.begin_generate()` and `FirestoreCandidateGenerationOperationRepository.begin_generate()` now inspect the candidate collection while the operation is pending. When a document has matching request metadata, they mutate the operation to `completed` and return it as a successful replay.

That state is not a legitimate partial commit under this PR's own contract: candidate creation and the transition to `completed` are one atomic transaction. If the operation is still pending, an existing candidate document is an impossible/corrupt state, not evidence that completion partially succeeded. The implementation comment says the two “must never coexist,” but then adopts the document.

This also allows a pre-seeded or recreated candidate that copies the request metadata to become canonical without passing through the operation's completion transaction.

**Required:** treat any pending-operation-plus-candidate state as `GenerateOperationIntegrityError` and leave both records untouched. Invert the new `test_generate_ops_begin_reconciles_pending_with_existing_owned_candidate` expectation. A legacy repair/migration path would be a separate, explicitly authorized capability.

### 2. The first candidate write does not prove that the candidate belongs to the operation

In both repository implementations, `complete_generate()` validates `candidate_id`, but `candidate_belongs_to_generate_operation()` is only called when a candidate document already exists or the operation is already completed. On the normal first-write path, the candidate is created and the operation is completed without validating its generation receipt against the request ID, caller scope, and request digest.

A malformed or incorrectly assembled candidate can therefore be durably committed and returned successfully on the first call, then fail integrity checks on the first replay. That breaks the stable-outcome guarantee.

**Required:** validate the candidate-operation binding before every candidate create/commit, not only adoption/replay. Add memory and Firestore tests that complete an empty reservation using the correct candidate ID but a mismatched request ID, caller scope, missing digest, and wrong digest; every case must fail without creating a candidate or changing the operation.

### 3. Request-digest binding still cannot detect recreation of a different candidate for the same input

`candidate_belongs_to_generate_operation()` proves only:

- candidate ID;
- request ID;
- caller scope;
- input request digest.

Those fields bind the candidate to the generation request, but not to the exact generated outcome. A deleted/recreated document can copy that receipt metadata while replacing the definition, validation receipt, assets, warnings, or other response fields, and completed replay will accept the replacement as canonical.

This conflicts with the handoff's “replay returns the exact candidate” and “deletion/recreation fails closed” requirements.

**Required:** atomically retain a candidate outcome fingerprint on completion and verify it on replay, existing-document checks, and indeterminate reconciliation. Prefer a canonical digest of the complete replayable candidate envelope; at minimum, the operation must bind the immutable mechanics/output rather than only the input request. Add a test that replaces the candidate under the same ID while preserving request ID/scope/request digest but changing the definition, and verify replay fails closed.

### 4. An indeterminate failure write does not recover a concurrently completed success

`FirestoreCandidateGenerationOperationRepository.fail_generate()` re-reads after an uncertain transaction, but only returns when the durable state is `failed`. If another worker completed successfully, it raises `TransactionIndeterminateError`. The service catches that together with `PersistenceUnavailableError` and returns a generic persistence failure instead of invoking terminal-state reconciliation.

The completion path now correctly re-reads and can return a durable success after an indeterminate transaction; the failure path needs the same behavior. The handoff explicitly requires returning a completed candidate when the operation and candidate prove that commit.

**Required:** route `TransactionIndeterminateError` from `fail_generate()` through `_resolve_after_terminal_race(..., indeterminate=True)` using the pinned digest and reserved candidate ID. Add a test where the local provider result is failure, the failure transaction becomes indeterminate, and the durable operation has concurrently completed; the caller must receive the canonical candidate rather than `persistence_unavailable`.

### Verification and handback

The current `Red-team hardening` run for this head is still failing in the test step. I could identify the failed job, but the available log output did not expose the final assertion, so it remains unclassified as baseline or PR-caused.

The PR body is also still stale: it records the original head `0b61558d...`, the original OpenAPI fingerprint, and the original test count despite four commits and 23 changed files. Update the required handback and account for all final paths and verification results before merge.

Drakosfire (commented):
Re-review of `90ad0703140b87624a4ece54295beacca910d47e`.

The correction fully resolves two prior findings: first-write candidate ownership is now checked before persistence, and an indeterminate failure write can reconcile a concurrently completed success. The new operation outcome fingerprint is also the right direction.

**Verdict: still not ready to merge. Three blockers remain.**

### 1. `pending + candidate exists` still becomes success through `complete_generate`

`begin_generate()` now correctly treats a pending operation plus an existing candidate document as an impossible atomic state. However, both completion repositories still accept the same state:

- Firestore enters `if existing.status is completed or candidate_snap.exists`, validates the pre-existing candidate against the pending operation, computes an outcome digest, and promotes the operation to completed.
- The memory repository catches candidate-create conflict, reloads the pre-existing candidate, validates it, and then promotes the pending operation to completed.

Because candidate creation and operation completion are one atomic unit, a candidate cannot legitimately exist while the operation remains pending. Same-operation stale-worker convergence should occur only after the operation is already completed. While pending, any existing candidate—even one with copied receipt metadata—must fail closed.

The new regression test covers only a subsequent `begin_generate()` after pre-seeding a candidate; it does not call `complete_generate()` against that state.

**Required:** in both repositories, when the operation is pending and the candidate document already exists, raise `GenerateOperationIntegrityError` without mutating the operation. Add direct memory and Firestore completion tests for a pre-existing candidate whose receipt matches the pending operation.

### 2. `outcome_digest` does not bind the exact candidate returned by replay

`compute_candidate_outcome_digest()` covers only:

- canonical definition;
- asset references;
- asset warnings.

The replayable candidate also contains `validation_receipt`, `generation_receipt`, `asset_brief`, `created_at`, `expires_at`, and `source_locator`. A recreated document can preserve request receipt metadata plus definition/assets/warnings, alter any of those other fields, and still pass ownership and outcome verification. This is especially visible for `expires_at`: the operation controls whether replay is allowed, but before operation expiry the server can return a candidate document advertising a different expiry than the original.

That does not satisfy the handoff's “exact persisted candidate” or deletion/recreation fail-closed guarantees.

**Required:** bind the canonical full replay response, excluding only fields that are explicitly and equivalently anchored elsewhere. At minimum verify every persisted candidate field returned by the generate POST, and require candidate `expires_at` to equal the operation's recorded `candidate_expires_at`. Add replacement tests for `asset_brief`, validation/provenance metadata, source locator, and candidate expiry—not only altered definition mechanics.

### 3. Completion reconciliation bypasses authoritative expiry and missing-candidate semantics

Ordinary completed replay goes through `_load_completed_candidate()`, which applies operation-recorded expiry and distinguishes premature loss from expected expiry. Repository completion and `_reconcile_complete()` return the stored candidate directly instead.

Consequences:

- a stale or indeterminate completing worker can return HTTP success after `candidate_expires_at` when the TTL-delayed candidate document still exists;
- if a completed operation's candidate is definitely absent during `_reconcile_complete()`, the repository raises `PersistenceUnavailableError` rather than `CandidateMissingBeforeExpiryError` before expiry or `CandidateExpiredError` afterward.

The handoff requires transaction-indeterminate recovery to establish and return the same canonical completed outcome semantics as an ordinary replay.

**Required:** route every discovered completed state through one authoritative completed-candidate loader, or return the completed operation to the service and apply the same ownership, outcome-digest, expiry, and missing-candidate rules there. Add reconciliation tests for retained-but-expired candidates and missing candidates both before and after the recorded expiry.

### CI and handback

The current `Red-team hardening` workflow is still failing in its test step. The available connector log remains truncated before the final assertion, so the failure is still unclassified.

The PR body also remains stale: it reports the original head `0b61558d...`, original fingerprint, and original test count despite the PR now containing five commits and substantially more tests. Update the required handback before merge.

Drakosfire (commented):
Re-review of `2cb566b0c5a36f29077a43c154d723a747a3aa1f`.

The commit closes the three previously reported paths: pending+candidate now fails closed in begin/complete/reconciliation, `outcome_digest` covers the full candidate payload, and successful completion is routed through the authoritative completed-replay loader.

The PR is still not ready to merge. Remaining blockers:

1. **Generate idempotency still fails open when the operation repository is absent.** `GenerationServiceV1.generate()` directly executes `_run()` whenever `generate_operations is None`, so repeated identical `request_id` values invoke the provider and allocate new candidates. `build_generation_service()` explicitly supports a `client=None` composition with injected candidate/persistence repositories and leaves `_generate_operations` unset; the production-composition test asserts this state. The generate contract must fail closed when durable operations are unavailable. Make the operation repository required for `generate`, or return a typed unavailable/misconfigured result; do not preserve the old non-idempotent fallback. Revision can remain independent.

2. **The lease budget ignores the optional asset-generation phase.** Configuration validates only `provider_timeout * attempts + 30`, while `_run()` invokes the asset gateway after provider completion and the gateway has an independently configurable timeout. An accepted configuration with `STATBLOCKS_V1_ASSET_TIMEOUT_SECONDS > 30` can therefore expire the lease during an ordinary request, allowing takeover and duplicate provider/asset work before the first worker finishes. Include enabled asset timeout in both the default and minimum accepted lease budget, or otherwise prove/constrain the full normal operation duration. Add a configuration test with an asset timeout larger than the current fixed margin.

3. **The revise OpenAPI contract now advertises generate-idempotency failures.** `_REVISE_ERROR_RESPONSES` spreads `_CANDIDATE_ERROR_RESPONSES`, which now includes `409 idempotency conflict/generation in progress`, `410 completed generation replay expired`, and generate-operation integrity text. The handoff explicitly requires candidate revise to remain non-idempotent and unchanged. Split generate-only responses from the shared provider/validation responses, regenerate OpenAPI/client artifacts, and assert the revise operation does not expose generate replay statuses.

4. **The post-completion reread is not bound to the original reservation.** `_outcome_from_completed_generate()` receives only `caller_scope`, `request_id`, and `replayed`; unlike `_resolve_after_terminal_race()`, it does not verify that the reloaded completed record still has the claimed `candidate_id` and pinned `request_digest`. It then validates the candidate against whatever record was reloaded, so a mismatched/replaced record can redirect a successful call to a different operation outcome instead of failing closed. Pass and verify the original reserved candidate ID and digest before returning the reloaded result; add a success-path mismatch regression.

5. **Operation expiry is not cross-checked against the bound candidate expiry.** `outcome_digest` now includes the candidate's `expires_at`, but `candidate_expires_at` remains an independent operation field. `_load_completed_candidate()` trusts the operation timestamp before loading and never requires `candidate.expires_at == operation.candidate_expires_at`. A malformed operation with a later timestamp can return a candidate after its own declared expiry; an earlier timestamp can produce a false normal 410. Treat disagreement as `GenerateOperationIntegrityError` and test both directions. For retained documents, verify ownership/outcome/expiry equality before applying the authoritative expiry decision; missing documents can continue using the operation timestamp to distinguish premature loss from TTL expiry.

CI/handback:
- The current `Red-team hardening` run still fails in `Run red-team hardening tests`; the connector log is truncated before the final assertion, so I could not classify the failure.
- The PR body still reports the original head, fingerprint, and 219-test result despite six commits and substantial additional tests. Refresh the required handback, exact current test commands/results, OpenAPI fingerprint, path exceptions, and live-proof evidence before merge.

Drakosfire (commented):
Reviewed the correction commit against the five findings from the prior review.

Four are resolved: generate now fails closed without durable operation persistence; production lease validation includes configured asset timeout; revise no longer advertises generate-only replay errors; and successful completion rereads are bound to the original candidate reservation and request digest.

One blocker remains:

**[P1] A corrupted earlier operation expiry still produces a false 410 before expiry binding can be checked.**

`_load_completed_candidate()` checks `now >= operation.candidate_expires_at` and raises `CandidateExpiredError` before loading the retained candidate. The new `candidate.expires_at == operation.candidate_expires_at` integrity check therefore only runs while the operation timestamp is still in the future.

Concrete case: the persisted candidate expires at 12:00, the operation record is corrupted to 11:00, and replay occurs at 11:30. The candidate is present and unexpired, but this path returns ordinary 410 instead of the documented integrity failure. The new test covers the opposite shape—both timestamps still in the future—so it does not exercise this ordering hole.

The documented contract says that a present candidate whose expiry disagrees with the operation is an integrity failure. Load without candidate-TTL enforcement first when possible; when the candidate exists, validate ownership/outcome binding and expiry equality before applying 410. When the candidate is absent, use the operation timestamp to distinguish normal post-expiry deletion from premature loss. Add tests with `now` between the mismatched timestamps in both directions.

CI remains red: the current `Red-team hardening` run fails in its test step. The connector log is truncated before the final assertion, so I could not classify the failure.

The PR body also remains stale: it still reports the original head, fingerprint, lease formula, and 219-test result despite seven commits and 24 changed files.

Drakosfire (commented):
Re-review of `7a214741af81d3895866c5f29332937bd2af87ab`.

The final code blocker from the prior review is closed. Completed replay now loads the retained candidate before applying ordinary expiry, verifies candidate ownership and the full outcome digest, verifies candidate/operation expiry agreement, and only then returns 410. A missing candidate after the durable operation expiry remains an ordinary 410, while a present candidate with a mismatched expiry fails closed as integrity. The added tests cover both the corrupted-past-operation-expiry case and the legitimate missing-after-expiry case.

I found no new code blocker in this commit. **Approved on source review.** GitHub does not permit this account to submit an `APPROVE` review on its own pull request, so this conclusion is recorded as a commit-anchored comment instead.

Two merge-readiness items remain outside this code finding:

1. The current `Red-team hardening` workflow still fails in `Run red-team hardening tests`. The available connector log is truncated before the final assertion, so the failure remains unclassified and should be resolved or explicitly shown to be inherited before merge.
2. The PR body is still stale: it reports the original head, old lease formula, old OpenAPI fingerprint/test count, and incomplete outside-allowlist disclosure. Update the required handback to the actual final branch state before merge.

I was not able to run the repository test suite independently in this environment.

PR #25, opened 2026-07-29
feat(statblocks_v1): emit durable bounded diagnostics for definition_invalid

Drakosfire:
## Summary

Implements **DMS-VAL-01** per DungeonMindBuddy's checked-in handoff `HANDOFF-dms-generation-validation-diagnostics` (PR Drakosfire/DungeonMindBuddy#449). Parent evidence: R0-A Mireward Latchling dogfood `FAIL_PRODUCT` — a real generation failure collapsed to `Generated definition failed validation`, leaving the operator with no actionable cause.

Generated definitions rejected before candidate creation now emit a **bounded typed diagnostic packet** on HTTP 422 `validation_failed`, persisted on the terminal failure snapshot so same-key replay exposes the identical packet with no second provider call.

## Diagnostic packet contract

`GenerationValidationDiagnosticPacketV1`: `schema_version` discriminator, `phase` (`schema_validation` | `domain_validation`), `issue_count`, and bounded `issues[]` (≤32 entries; `field_path`≤256, `code`≤64, `message`≤512, `suggested_resolution`≤512), deterministic ordering.

- **schema_validation** derives from Pydantic `ValidationError` via `errors(include_input=False, include_url=False)`; severity=`error`; `suggested_resolution` omitted; stable UPPER_SNAKE codes normalized from pydantic error types.
- **domain_validation** derives from error-severity `ValidationReceiptV1.issues`.
- Normalization failure **fails closed**: `diagnostics=None`, generic public message retained.
- Packet persisted as optional field on `CandidateGenerationFailureSnapshotV1` — **legacy records without it load and replay generically**.
- **Replay derives the packet from the persisted snapshot**; provider call count unchanged; changed-digest replay remains 409.
- `raise_for_generation_failure` attaches the packet to `error.details` only when present; unknown failure kinds still fail closed with no details.

## Security

Raw provider payload, `ValidationError.input`, provider exception objects, and prompts are never serialized into responses, logs, or durable operation records. Proven by sentinel test (`__DMS_VAL01_RAW_PAYLOAD_SENTINEL__`) across diagnostics JSON, operation record JSON, and HTTP response text, plus `'"input":'` / `'"ctx":'` absence assertions.

## Verification (all re-run by reviewer, not just the implementer)

| Gate | Result |
|---|---|
| `uv run pytest tests/statblocks_v1 -q` (project venv, lock pydantic 2.10.6) | **276 passed, 15 skipped, 0 failed** |
| `./scripts/run_statblocks_v1_tests.sh -q` (isolated lane) | **0 failures** (lane repinned to lock pydantic on main) |
| Firestore integration | 2 passed, 13 skipped — emulator not configured in this environment (honest skips, incl. new diagnostics round-trip test) |
| `git diff --check main..HEAD` | clean |

Key proofs: first-vs-replay packet equality + one provider call (`test_definition_invalid_failure_replay_preserves_diagnostics_and_one_provider_call`); generate **and** revise share the envelope contract (`test_definition_invalid_generate_and_revise_emit_matching_diagnostic_envelopes`); legacy snapshot without diagnostics replays generically; memory + Firestore round-trips.

## Stacks on (already on main)

- `01abf2f` schema artifact regeneration (fail-closed blocker for clean checkouts)
- `6842cb8` OpenAPI contract artifact regeneration (same emission-drift class; TS client byte-identical except fingerprint)
- `8aff6ea` isolated-lane pydantic repin

## Explicitly not done (per handoff §5)

No repair retry, no second provider call, no validation relaxation, no prompt tuning, no schema/compiler changes, no DungeonMindBuddy code. Buddy propagation + Workbench rendering is the named successor (**BUDDY-VAL-01**).

## Reviewer protocol (handoff §10)

1. Reproduce both schema and domain invalidity — `test_schema_invalid_definition_emits_bounded_schema_diagnostics`, `test_domain_invalid_definition_emits_validator_issues`.
2. Inspect the persisted operation record, not only the HTTP body — replay test asserts `operation.failure.diagnostics` equals response packet.
3. Replay exact request, compare semantic packet equality; confirm one provider call.
4. Search response/logs/record for sentinel raw input.
5. Load a legacy failed record without diagnostics.
6. Confirm revise shares the invariant.

### DIFF
--- a/statblocks_v1/api/http_errors.py
+++ b/statblocks_v1/api/http_errors.py
@@ -134,7 +134,7 @@ def raise_for_generation_failure(failure: GenerationFailureV1) -> None:
         (500, "generation_failed", "Generation failed with an unexpected outcome"),
     )
     details: dict[str, object] | None = None
-    if failure.diagnostics is not None:
+    if failure.kind == "definition_invalid" and failure.diagnostics is not None:
         details = failure.diagnostics.model_dump(mode="json", exclude_none=True)
     raise StatblockV1HTTPError(
         status,
--- a/statblocks_v1/api/http_errors.py
+++ b/statblocks_v1/api/http_errors.py
@@ -133,7 +133,13 @@ def raise_for_generation_failure(failure: GenerationFailureV1) -> None:
         failure.kind,
         (500, "generation_failed", "Generation failed with an unexpected outcome"),
     )
-    raise StatblockV1HTTPError(status, GenerationTransportError(code, message))
+    details: dict[str, object] | None = None
+    if failure.diagnostics is not None:
+        details = failure.diagnostics.model_dump(mode="json", exclude_none=True)
+    raise StatblockV1HTTPError(
+        status,
+        GenerationTransportError(code, message, details=details),
+    )
 
 
 def register_error_handlers(app: FastAPI) -> None:
--- a/statblocks_v1/application/generation.py
+++ b/statblocks_v1/application/generation.py
@@ -1358,6 +1358,134 @@ def _pydantic_error_message(error_type: str) -> str:
     return _SCHEMA_DIAGNOSTIC_MESSAGES.get(error_type, _GENERIC_SCHEMA_DIAGNOSTIC_MESSAGE)
 
 
+_GENERIC_DOMAIN_DIAGNOSTIC_MESSAGE = "Domain validation failed for this field."
+
+_DOMAIN_DIAGNOSTIC_MESSAGES: dict[str, str] = {
+    "DEFAULT_ARMOR_CLASS_CARDINALITY": (
+        "Exactly one armor-class profile must be marked default."
+    ),
+    "HP_METHOD_FIELDS_INCOHERENT": (
+        "Hit point method fields are incoherent for the selected method."
+    ),
+    "HP_DISPLAYED_AVERAGE_MISMATCH": (
+        "Displayed HP average does not match the typed dice formula."
+    ),
+    "RULESET_CR_INVALID": "Challenge rating is not a canonical D&D 5e CR value.",
+    "RULESET_CR_PROFICIENCY_MISMATCH": (
+        "Challenge proficiency bonus does not match the selected CR."
+    ),
+    "PASSIVE_PERCEPTION_MISMATCH": (
+        "Passive Perception must equal 10 + Perception skill value."
+    ),
+    "PASSIVE_PERCEPTION_UNVERIFIED": (
+        "Passive Perception differs from Wisdom without a typed Perception skill."
+    ),
+    "DUPLICATE_SAVING_THROW_ABILITY": (
+        "Saving throw for the same ability appears more than once."
+    ),
+    "SAVING_THROW_DERIVATION_MISMATCH": (
+        "Declared saving throw value does not match the expected derivation."
+    ),
+    "DUPLICATE_SKILL_NAME": (
+        "Skill name duplicates an earlier entry after normalization."
+    ),
+    "SKILL_DERIVATION_MISMATCH": (
+        "Declared skill value does not match the expected derivation for the "
+        "linked ability."
+    ),
+    "DUPLICATE_LOCAL_KEY": "Local key is duplicated within the indicated collection.",
+    "DEFAULT_PHASE_CARDINALITY": (
+        "A phased creature must have exactly one default phase."
+    ),
+    "UNKNOWN_ELEMENT_REFERENCE": "Referenced element key does not exist.",
+    "UNKNOWN_RESOURCE_REFERENCE": "Referenced resource key does not exist.",
+    "UNKNOWN_MULTIATTACK_ELEMENT": (
+        "Multiattack references an element key that does not exist."
+    ),
+    "FORBIDDEN_REFERENCE_CYCLE": (
+        "Multiattack reference would create a forbidden cycle."
+    ),
+    "UNKNOWN_PHASE_REFERENCE": "Referenced phase key does not exist.",
+    "UNKNOWN_MOVEMENT_REFERENCE": "Referenced movement mode key does not exist.",
+    "UNKNOWN_PHASE_ELEMENT": "Phase references an element key that does not exist.",
+    "PHASE_ELEMENT_SET_CONFLICT": (
+        "Element cannot be both enabled and disabled in the same phase."
+    ),
+    "SECTION_ACTIVATION_INCOHERENT": (
+        "Activation kind is not valid for the element section."
+    ),
+    "REACTION_TRIGGER_REQUIRED": (
+        "A reaction needs a trigger or timing expression."
+    ),
+    "LEGENDARY_RESOURCE_REQUIRED": (
+        "Legendary actions require valid resource usage and cost entries."
+    ),
+    "LEGENDARY_RESOURCE_MISMATCH": (
+        "Legendary usage resource key must match every cost resource key."
+    ),
+    "LAIR_CONTEXT_REQUIRED": "A lair action requires a lair profile.",
+    "LAIR_TIMING_REQUIRED": (
+        "A lair action requires an initiative timing expression."
+    ),
+    "RESOURCE_COST_DUPLICATE_POOL": (
+        "Resource pool appears more than once in costs; use a single cost entry."
+    ),
+    "RESOURCE_COST_EXCEEDS_POOL": "Cost exceeds the resource pool maximum.",
+    "HUMAN_ADJUDICATED_AUTOMATION_MISMATCH": (
+        "Human-adjudicated mechanics must declare manual automation support."
+    ),
+    "USAGE_FIELDS_INCOHERENT": (
+        "Usage fields are incoherent for the selected usage kind; check the "
+        "requirements for that kind at the indicated field."
+    ),
+    "ATTACK_REACH_REQUIRED": "A melee attack requires typed reach.",
+    "ATTACK_RANGE_UNEXPECTED": "A melee attack must not include typed range.",
+    "ATTACK_RANGE_REQUIRED": "A ranged attack requires typed range.",
+    "ATTACK_REACH_UNEXPECTED": "A ranged attack must not include typed reach.",
+    "ATTACK_RANGE_ORDER_INCOHERENT": (
+        "Attack long range must use the same unit and be >= normal range."
+    ),
+    "ATTACK_TARGET_RANGE_UNEXPECTED": (
+        "Attack target must not set range; use mechanic.reach or mechanic.range."
+    ),
+    "ATTACK_TARGET_COUNT_REQUIRED": "A creatures target requires count.",
+    "ATTACK_TARGET_COUNT_INCOHERENT": (
+        "Target count is incoherent for the selected target kind."
+    ),
+    "ATTACK_TARGET_AREA_REQUIRED": "An area target requires area.",
+    "ATTACK_TARGET_AREA_UNEXPECTED": "Only an area target may set area.",
+    "SPELLCASTING_MODE_INCOHERENT": (
+        "Required spellcasting fields are missing for the selected casting mode."
+    ),
+    "SPELL_GROUP_SLOTS_INCOHERENT": (
+        "Spell group slots are incoherent for the spellcasting mode and group level."
+    ),
+    "SPELL_GROUP_LEVEL_INCOHERENT": (
+        "Spell group level is missing or outside the allowed range."
+    ),
+    "SPELL_GROUP_USAGE_INCOHERENT": (
+        "Spell group usage kind is incoherent for the spellcasting mode and group "
+        "level."
+    ),
+    "RULES_TEXT_ATTACK_BONUS_MISMATCH": (
+        "Rules text attack bonus conflicts with typed attack_bonus."
+    ),
+    "RULES_TEXT_DAMAGE_MISMATCH": (
+        "Rules text damage conflicts with the first typed hit damage effect."
+    ),
+    "RULES_TEXT_SAVE_DC_MISMATCH": (
+        "Rules text save DC conflicts with typed save DC."
+    ),
+    "RULES_TEXT_SECTION_MISMATCH": (
+        "Reaction rules text unambiguously says it is used as an action."
+    ),
+}
+
+
+def _domain_diagnostic_message(code: str) -> str:
+    return _DOMAIN_DIAGNOSTIC_MESSAGES.get(code, _GENERIC_DOMAIN_DIAGNOSTIC_MESSAGE)
+
+
 _UNEXPECTED_PROVIDER_KEY_FIELD = "<unexpected_key>"
 
 
@@ -1443,17 +1571,10 @@ def _diagnostics_from_domain_receipt(
                         issue.field_path,
                         max_len=MAX_GENERATION_VALIDATION_FIELD_PATH_LEN,
                     ),
-                    message=_bound_public_text(
-                        issue.message, max_len=MAX_GENERATION_VALIDATION_MESSAGE_LEN
-                    ),
-                    suggested_resolution=(
-                        _bound_public_text(
-                            issue.suggested_resolution,
-                            max_len=MAX_GENERATION_VALIDATION_SUGGESTED_RESOLUTION_LEN,
-                        )
-                        if issue.suggested_resolution is not None
-                        else None
-                    ),
+                    message=_domain_diagnostic_message(issue.code),
+                    # Error-severity issues never carry resolutions (only warnings do,
+                    # filtered above); packets must not reflect receipt text.
+                    suggested_resolution=None,
                 )
             )
         issues = sorted(
--- a/statblocks_v1/application/generation.py
+++ b/statblocks_v1/application/generation.py
@@ -1318,7 +1318,28 @@ def _pydantic_error_code(error_type: str) -> str:
     return normalized[:MAX_GENERATION_VALIDATION_CODE_LEN]
 
 
-def _field_path_from_pydantic_loc(loc: tuple[object, ...]) -> str:
+_UNEXPECTED_PROVIDER_KEY_FIELD = "<unexpected_key>"
+
+
+def _field_path_from_pydantic_loc(
+    loc: tuple[object, ...], *, error_type: str = ""
+) -> str:
+    if error_type == "extra_forbidden" and loc:
+        parent_loc = loc[:-1]
+        if not parent_loc:
+            path = _UNEXPECTED_PROVIDER_KEY_FIELD
+        else:
+            path = (
+                f"{_field_path_segments_to_public_path(parent_loc)}"
+                f".{_UNEXPECTED_PROVIDER_KEY_FIELD}"
+            )
+        return _bound_public_text(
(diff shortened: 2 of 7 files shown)


statblocks_v1/application/generation.py
```diff
@@ -1290,6 +1300,133 @@ def _with_key_preservation_warnings(
...
+        return "$"
+    segments: list[str] = []
+    for part in loc:
+        if isinstance(part, int):
+            if not segments:
+                segments.append(f"[{part}]")
+            else:
+                segments[-1] = f"{segments[-1]}[{part}]"
+        else:
+            segments.append(str(part))
```
Drakosfire, line 1521: **Blocking — raw provider keys leak into the durable/public packet.** With `extra="forbid"`, Pydantic reports an unexpected object key as part of `loc` (for example, payload key `__RAW_KEY_SENTINEL__` produces `loc=("__RAW_KEY_SENTINEL__",)`). This line serializes that provider-controlled key verbatim into `field_path`, which is then persisted and returned. The current sentinel test puts the sentinel in a value, so it cannot detect this path. Please sanitize/replace dynamic `loc` segments—at minimum the terminal segment for `extra_forbidden`—and add a raw-key sentinel test across first response, operation JSON, and replay.

statblocks_v1/api/http_errors.py
```diff
@@ -133,7 +133,13 @@ def raise_for_generation_failure(failure: GenerationFailureV1) -> None:
         failure.kind,
         (500, "generation_failed", "Generation failed with an unexpected outcome"),
     )
-    raise StatblockV1HTTPError(status, GenerationTransportError(code, message))
+    details: dict[str, object] | None = None
+    if failure.diagnostics is not None:
```
Drakosfire: The handoff and PR body require unknown failures to remain fail-closed with no details, but this condition emits any attached packet regardless of `failure.kind`. `GenerationFailureV1` currently permits diagnostics on every kind, so accidental/future construction of an unknown or provider failure with diagnostics would expose them. Gate this on `failure.kind == "definition_invalid"` (and consider enforcing the same invariant when creating/persisting snapshots), with a test proving an unknown kind carrying diagnostics still returns generic `generation_failed` without `details`.

Drakosfire (commented):
**Review verdict: changes required before merge.**

The durable first-response/replay design is sound: the packet is derived at the validation boundary, committed with the terminal operation snapshot, replayed without another provider call, and legacy records remain readable. The focused tests cover both schema/domain phases, generate/revise parity, memory persistence, and the Firestore path.

Two contract violations remain:

1. Schema diagnostics can currently serialize raw provider-supplied object keys through Pydantic `loc`, contradicting the raw-payload non-retention invariant. Add a key-sentinel adversarial test covering the diagnostics packet, persisted operation, first HTTP response, and same-key replay, then sanitize or replace dynamic `loc` segments—especially `extra_forbidden`—while retaining an actionable parent path.
2. The HTTP boundary emits `diagnostics` for any failure kind whenever the field is populated. Gate details to `definition_invalid` (and ideally enforce that invariant in the failure/snapshot model) so unknown and non-validation failures remain generic even under accidental construction. Add an unknown-kind-with-diagnostics test.

The failing `Red-team hardening` workflow is the known unrelated import-time `OpenAI()` collection failure, not introduced by this diff. After the two fixes, rerun the statblocks-v1 suite and the first-failure → same-key replay proofs.

Drakosfire:
Both blocking findings addressed in `0dcc5cf`, independently re-verified by the reviewer (not just the implementer).

## Finding 1 — raw provider keys in `field_path` — FIXED

Confirmed before fixing: `extra_forbidden` `loc` ends with the raw provider key (reproduced with `__RAW_PROVIDER_KEY_SENTINEL__`). Also scoped the blast radius: **no `dict[...]` fields exist anywhere in the StatblockDefinitionV1 model graph**, so the `extra_forbidden` terminal segment is the only provider-controlled `loc` channel — the fix is targeted, not a blunt regex.

Fix (`generation.py`): for `extra_forbidden` only, the terminal `loc` segment is dropped and replaced with fixed placeholder `<unexpected_key>`, preserving the parent path:
- nested `('rule_elements', 0, key)` → `rule_elements[0].<unexpected_key>`
- root-level `(key,)` → `<unexpected_key>`

Raw-key sentinel coverage (all four required surfaces):
- diagnostics JSON — `test_extra_forbidden_provider_key_sentinel_sanitized_at_service_and_persistence` (also asserts parent preservation and `EXTRA_FORBIDDEN` code normalization)
- persisted operation JSON — same test
- first HTTP response + same-key replay response — `test_extra_forbidden_provider_key_sentinel_absent_from_generate_http_and_replay` (also asserts one provider call)

Reviewer repro after fix: sentinel absent from packet JSON; `EXTRA_FORBIDDEN -> <unexpected_key>`.

## Finding 2 — fail-closed at HTTP boundary — FIXED

- `raise_for_generation_failure` now attaches `details` only when `failure.kind == "definition_invalid" and failure.diagnostics is not None`.
- `CandidateGenerationFailureSnapshotV1` additionally rejects `diagnostics` on any other kind at construction (persistence boundary) — misbound packets can no longer be persisted or replayed; legacy records and `definition_invalid` ± packet still parse.
- Required test added: unknown kind deliberately carrying a packet → **500 `generation_failed`, no `details`** (`test_misbound_generation_failure_never_exposes_diagnostics_in_http`), plus direct policy coverage in `test_raise_for_generation_failure_omits_diagnostics_unless_definition_invalid` and the snapshot invariant in `test_candidate_generation_failure_snapshot_diagnostics_invariant`.

## Gate

`uv run pytest tests/statblocks_v1 -q` → **281 passed, 15 skipped, 0 failed** (re-run by reviewer); `git diff --check` clean.

statblocks_v1/application/generation.py
```diff
@@ -1290,6 +1300,158 @@ def _with_key_preservation_warnings(
...
+            raw_issues.append(
+                GenerationValidationDiagnosticIssueV1(
+                    code=_pydantic_error_code(error_type),
+                    severity=ValidationSeverity.error,
+                    field_path=_field_path_from_pydantic_loc(
+                        tuple(item.get("loc", ())),
+                        error_type=error_type,
+                    ),
+                    message=_bound_public_text(
+                        str(item.get("msg", "Validation failed")),
```
Drakosfire: **Blocking — rejected provider values can still leak through Pydantic `msg`.** The statblock contract has discriminated unions (`Mechanic`, and nested effects). For example, setting `rule_elements[0].mechanic.kind` to `__RAW_PROVIDER_VALUE_SENTINEL__` produces a `union_tag_invalid` error whose `msg` includes that exact tag. This line persists and returns the interpolated message verbatim; `include_input=False`/omitting `ctx` does not sanitize it. Map schema messages to stable trusted templates by `error_type` (with unknown types generic), and add a value-sentinel test covering packet JSON, operation JSON, first response, replay, and one provider call.

Drakosfire (commented):
**Re-review verdict: the two original findings are fixed, but one adjacent privacy blocker remains.**

`0dcc5cf` correctly replaces `extra_forbidden` key segments, preserves the parent path, gates HTTP details to `definition_invalid`, and rejects misbound packets at the persisted snapshot boundary. The added first-response/replay and invariant tests are well targeted.

The schema packet still reflects Pydantic's interpolated `msg` verbatim. For the actual discriminated `Mechanic` union, a provider value such as `rule_elements[0].mechanic.kind = "__RAW_PROVIDER_VALUE_SENTINEL__"` produces `union_tag_invalid`, whose message includes that rejected tag value. `include_input=False` and omitting `ctx` do not remove it because the value is already interpolated into `msg`.

Please normalize schema messages to trusted templates keyed by error type—or at minimum handle all Pydantic error kinds that interpolate input, with unknown kinds failing to a generic message. Add a value-sentinel test across diagnostics JSON, persisted operation JSON, first HTTP response, and same-key replay; confirm one provider call.

The new Actions failure remains the same unrelated import-time `OpenAI()` baseline. I found no regression in the two fixes themselves.

Drakosfire:
## Closure: provider values in Pydantic messages (commit 4e47a32)

### Root cause confirmed

Reproduced before fixing: `mechanic.kind = "__RAW_PROVIDER_VALUE_SENTINEL__"` produces Pydantic error type `union_tag_invalid` with `msg`:

> `Input tag '__RAW_PROVIDER_VALUE_SENTINEL__' found using 'kind' does not match any of the expected tags: 'attack', ...`

`exc.errors(include_url=False, include_input=False)` strips `input`/`ctx` but the provider value is already interpolated into `msg`, and the implementation copied `msg` into the durable packet. Real leak.

### Fix

Schema-phase diagnostic messages are now generated exclusively from trusted static templates keyed by Pydantic error type:

- `_SCHEMA_DIAGNOSTIC_MESSAGES` (`statblocks_v1/application/generation.py`): static DMS-authored strings for the error types this schema can produce — `missing`, `extra_forbidden`, `union_tag_invalid`, `union_tag_not_found`, `literal_error`, `enum`, the string/int/float/list/dict/model/bool structural and constraint types, and `value_error`. No `msg`, `ctx`, or `input` text is ever interpolated.
- `union_tag_invalid` → `Unrecognized element kind; the value does not match any supported kind for this field.` (no provider tag value)
- `union_tag_not_found` → `Missing the discriminator field required to determine the element type.`
- Unknown error types → generic `_GENERIC_SCHEMA_DIAGNOSTIC_MESSAGE` (`Validation failed for this field.`); Pydantic text is never reflected.

Post-fix probe of the same sentinel payload: `sentinel_in_packet: False`; issue is `code=UNION_TAG_INVALID`, `field_path=rule_elements[0].mechanic` (parent path preserved, still actionable), message is the template.

### Value-sentinel test coverage (all five surfaces)

- Diagnostic JSON: `test_union_tag_invalid_provider_value_sentinel_absent_from_diagnostics_and_persistence` (service) — sentinel absent from `result.diagnostics.model_dump_json()`; asserts the `UNION_TAG_INVALID` message equals the template constant.
- Durable operation JSON: same test — sentinel absent from persisted `operation.model_dump_json()`.
- First HTTP response: `test_union_tag_invalid_provider_value_sentinel_absent_from_generate_http_and_replay` (API) — sentinel absent from `first.text`.
- Same-key replay: same test — sentinel absent from `second.text`.
- One provider call: same test — `len(provider.calls) == 1`.
- `union_tag_not_found` sanitization: `test_union_tag_not_found_uses_template_message`.

### Verification

`uv run pytest tests/statblocks_v1 -q` → **284 passed, 15 skipped, 0 failed** (re-run independently after implementation). `git diff --check` clean. Diff touches only `generation.py`, `test_generation_service.py`, `test_candidate_routes.py` (124 insertions, 4 deletions).

### FYI — adjacent surface, not changed (flagging for adjudication)

Domain-phase diagnostic messages (`_diagnostics_from_domain_receipt`) originate from DMS's own validators in `domain/validation.py`, and a few interpolate provider-authored values — e.g. `Local key '{key}' is duplicated.`. This is materially narrower than the schema-phase leak: domain validation only runs on definitions that already passed schema validation, and the interpolated keys are pattern-constrained tokens (`key: str = Field(pattern=r"^[a-z][a-z0-9_]*$"`) — bounded, validated identifiers, not raw payload. I left this path unchanged per the review's scope; happy to harden it (template/redact key interpolation) in a follow-up if you want defense-in-depth there too.

statblocks_v1/application/generation.py
```diff
@@ -1290,6 +1300,195 @@ def _with_key_preservation_warnings(
...
+                continue
+            raw_issues.append(
+                GenerationValidationDiagnosticIssueV1(
+                    code=issue.code[:MAX_GENERATION_VALIDATION_CODE_LEN],
+                    severity=issue.severity,
+                    field_path=_bound_public_text(
+                        issue.field_path,
+                        max_len=MAX_GENERATION_VALIDATION_FIELD_PATH_LEN,
+                    ),
+                    message=_bound_public_text(
```
Drakosfire: **Blocking — domain diagnostics still reflect arbitrary provider-authored values.** This copies `ValidationIssueV1.message` verbatim. `domain/validation.py` emits error `DUPLICATE_SKILL_NAME` as `f"Skill '{skill.skill}' duplicates..."`, while `SkillBonus.skill` is only `Field(min_length=1)`—no pattern or maximum. Two identical sentinel skill names pass schema validation, trigger domain invalidity, and place the sentinel in this durable/public message. Normalize domain messages to trusted templates by issue code (or explicitly sanitize interpolated values), and add packet/operation/first-response/replay/one-call coverage.

Drakosfire (commented):
**Re-review verdict: schema-message leak fixed; one domain-phase privacy blocker remains.**

Commit `4e47a32` correctly stops reflecting Pydantic `msg`: schema messages now come from trusted templates, unknown error types use a generic fallback, and the union-tag value sentinel is absent from diagnostics, persistence, first response, and replay with one provider call.

However, the PR's raw-provider non-retention invariant covers the whole diagnostic packet, not only schema-phase failures. `_diagnostics_from_domain_receipt` still copies validator messages verbatim. At least one error-severity validator message contains arbitrary provider-authored prose: `DUPLICATE_SKILL_NAME` formats `skill.skill` into the message, while `SkillBonus.skill` is only `Field(min_length=1)` with no pattern or maximum length. A provider can submit the same `__RAW_DOMAIN_VALUE_SENTINEL__` skill twice, pass schema validation, fail domain validation, and have that value persisted and returned.

Please normalize domain diagnostic messages to trusted templates by issue code, or otherwise prove every interpolated field is safe and bounded. Add the same adversarial proof across packet JSON, operation JSON, first HTTP response, same-key replay, and one provider call. Constrained local-key interpolation can remain only if that exposure is an explicit contract decision rather than an accidental reflection.

The focused statblock suite is reported green (`284 passed, 15 skipped`). The Actions failure remains the unrelated import-time `OpenAI()` baseline.

Drakosfire:
## Closure: provider values in domain-validation messages (commit c3fbea1)

### Root cause confirmed

Reproduced exactly as described: `SkillBonus.skill` is `str = Field(min_length=1)` (no pattern, no max length). Two skills named `__RAW_DOMAIN_VALUE_SENTINEL__` pass schema validation, fail domain validation with `DUPLICATE_SKILL_NAME`, and `_diagnostics_from_domain_receipt()` copied the receipt message — `Skill '__RAW_DOMAIN_VALUE_SENTINEL__' duplicates an earlier entry after normalization.` — into the durable/public packet. My earlier "pattern-constrained tokens" FYI was wrong as a general defense; this counterexample defeats it.

### Fix (mirrors the schema-phase architecture)

Domain-phase packet messages now come exclusively from trusted static templates keyed by validator issue code:

- `_DOMAIN_DIAGNOSTIC_MESSAGES` (`statblocks_v1/application/generation.py`): a static, interpolation-free template for **all 49 codes** currently emitted by `domain/validation.py`, derived from the current validator text with provider interpolation removed (e.g. `DUPLICATE_SKILL_NAME` → `Skill name duplicates an earlier entry after normalization.`).
- Unknown/future codes → generic `_GENERIC_DOMAIN_DIAGNOSTIC_MESSAGE` (`Domain validation failed for this field.`); receipt text is never reflected.
- `suggested_resolution` in domain packets is now always `None`: error-severity issues never carry resolutions today (the only resolution-emitting issue is `PASSIVE_PERCEPTION_UNVERIFIED`, a warning, filtered out), and packets must not reflect receipt text.
- `field_path` handling unchanged: verified all domain field paths are structural (literal segments + integer indices) — no provider content.
- Validator messages in `domain/validation.py` are unchanged; only the packet boundary changed.

Post-fix probe of the same payload: `sentinel_in_domain_packet: False`; issue is `code=DUPLICATE_SKILL_NAME`, `field_path=proficiencies.skills[1].skill`, template message, `resolution=None`.

### Sentinel test coverage (all five surfaces)

- Diagnostic JSON: `test_domain_duplicate_skill_provider_value_sentinel_absent_from_diagnostics_and_persistence` — sentinel absent from `result.diagnostics.model_dump_json()`; message equals the template.
- Durable operation JSON: same test — sentinel absent from persisted `operation.model_dump_json()`.
- First HTTP response: `test_domain_provider_value_sentinel_absent_from_generate_http_and_replay` — sentinel absent from `first.text`.
- Same-key replay: same test — sentinel absent from `second.text`.
- One provider call: same test — `len(provider.calls) == 1`.
- Unknown-code fallback: `test_unknown_domain_code_uses_generic_message` — a fabricated-code receipt carrying hostile text (`Provider-authored receipt text must not leak.`) yields the generic message and `resolution=None`.
- Anti-drift guard: `test_domain_diagnostic_templates_cover_all_validator_codes` — parses `validation.py` and asserts its 49 emitted codes equal the template-map keys exactly, so adding a validator code without a template (or leaving a stale one) fails the suite.

### Trade-off (transparent)

Per-code templates collapse per-site message variants into one message per code (e.g. `USAGE_FIELDS_INCOHERENT` had ~15 site-specific messages). The packet still carries `code` + precise `field_path` (e.g. `rule_elements[3].usage.uses`), so the locator precision is preserved; the message is generic-per-code by design. This is the same trade the schema phase made for fail-closed guarantees.

### Verification

`uv run pytest tests/statblocks_v1 -q` → **288 passed, 15 skipped, 0 failed** (re-run independently after implementation). `git diff --check` clean. Diff touches only `generation.py`, `test_generation_service.py`, `test_candidate_routes.py` (263 insertions, 13 deletions).

### FYI — adjacent surface, not changed (flagging for adjudication)

The public `/statblock-definitions:validate` route (`editor_preview` mode) returns `ValidationReceiptV1` messages over HTTP, and those messages still interpolate provider-authored values (the same sentinel skill name appears in its `DUPLICATE_SKILL_NAME` message). That surface predates DMS-VAL-01 and is outside the durable-packet scope of this PR. If you want it hardened, the clean follow-up is de-interpolating validator messages at the source in `domain/validation.py` — which fixes every receipt surface at once and preserves per-site message specificity there. Happy to write that handoff.

Drakosfire (commented):
**Re-review verdict: no blocking findings remain in PR #25.**

Commit `c3fbea1` closes the domain-phase privacy gap correctly. Domain diagnostic messages now come exclusively from trusted templates keyed by internal validator code, unknown codes fail to a generic message, receipt text and suggested resolutions are not reflected, and the duplicate-skill sentinel is absent from the diagnostic packet, persisted operation, first HTTP response, and same-key replay with one provider call.

I re-checked the complete invariant chain across the PR:

- raw `extra_forbidden` provider keys are replaced while retaining an actionable parent path;
- schema messages never reflect Pydantic `msg`, `input`, or `ctx`;
- domain messages never reflect `ValidationReceiptV1.message`;
- unknown schema/domain error types use generic trusted text;
- diagnostics can only be persisted and emitted for `definition_invalid`;
- exact replay uses the persisted packet without another provider call;
- legacy failed records without diagnostics still load generically;
- generate and revise retain the shared envelope behavior.

The new template-coverage test also ensures every currently emitted validator code has an explicit domain message. The pre-existing `/statblock-definitions:validate` receipt surface remains outside this PR's durable-generation packet scope and can be tracked separately if broader receipt de-interpolation is desired.

The focused suite is reported at `288 passed, 15 skipped`. The failing `Red-team hardening` workflow remains the unrelated baseline import-time `OpenAI()` initialization caused by a missing `OPENAI_API_KEY`, not this diff.

GitHub does not permit approving one's own pull request through this account, but from the code-review perspective this is ready to merge once the known repository-level CI disposition is handled.

PR #27, opened 2026-07-30
feat(statblocks_v1): server-owned derived math (prompt v5)

Drakosfire:
## Summary

Stacked on #26. Answers "why are we asking an LLM to do arithmetic?" — we no longer do.

- New `statblocks_v1/domain/derived.py` owns the derivation math (CR→proficiency table, ability modifiers, standard/expertise expectations, HP average). `validation.py` now imports it instead of re-implementing, so check and compute can never drift apart.
- `GenerationServiceV1` computes server-owned values after provider parse, before validation/storage: skill/save `value` for standard/expertise (`explicit_override` keeps authored authority), `challenge.proficiency_bonus`, and `hit_points.displayed_average`. Provider numbers become advisory; overrides are logged as structured adjustments.
- `senses.passive_perception` deliberately stays authored (traits can legitimately raise it); validator semantics unchanged, so authored Workbench paths keep full error coverage.
- Prompt v5 deletes the arithmetic teaching (the derivation worked example + DERIVED MATH bullets) and replaces it with intent declaration — smaller prompt, smaller failure surface.
- Harness gains `--server-derived` to measure what production stores rather than raw model output.

## Evidence (thornvault-regent x5, arm both+ex, --server-derived)

- `gpt-5.4-mini`: `SKILL_DERIVATION_MISMATCH`, `SAVING_THROW_DERIVATION_MISMATCH`, `HP_DISPLAYED_AVERAGE_MISMATCH`, `RULESET_CR_PROFICIENCY_MISMATCH` all **0** (was 18+ derivation incidences across prior 10-trial runs on mini). The class is structurally dead regardless of model arithmetic skill.
- `gpt-5.6-luna`: 4/5 clean — matches the v4 confirmation run, no regression.
- Full lane: 299 passed, 16 skipped.

Known residuals (unchanged by this PR, tracked): mini's guidance-saturated `USAGE_FIELDS_INCOHERENT`/`SPELL_GROUP_SLOTS_INCOHERENT` clusters, and `PASSIVE_PERCEPTION_MISMATCH` is now more visible on mini since it stays authored (a candidate for the same treatment later).

## Test plan

- [x] `uv run pytest tests/statblocks_v1/ -q` (299 passed)
- [x] Harness mini + Luna runs with `--server-derived`
- [ ] After #26 merges: retarget this PR to `main`

### DIFF
--- a/statblocks_v1/application/prompts.py
+++ b/statblocks_v1/application/prompts.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 from statblocks_v1.application.commands import GenerateStatblockCommandV1, ReviseStatblockCommandV1
 from statblocks_v1.domain.rule_elements import StatblockDefinitionV1
 
-PROMPT_VERSION = "statblock-generation-prompt-v4"
+PROMPT_VERSION = "statblock-generation-prompt-v5"
 
 WORKED_EXAMPLES = """WORKED EXAMPLES - field shapes are normative; values are illustrative, compute yours per creature:
 - recharge usage: {"kind":"recharge","recharge_range":{"minimum":5,"maximum":6},"uses":null,"resource_key":null,"refresh_text":null}
@@ -12,7 +12,6 @@ WORKED_EXAMPLES = """WORKED EXAMPLES - field shapes are normative; values are il
 - resource usage spending from a declared pool: {"kind":"resource","recharge_range":null,"uses":null,"resource_key":"legendary_actions","refresh_text":null}
 - multiattack: mechanic.kind "multiattack" with "sequences":[{"element_key":"bite","count":1,"choice_group":null},{"element_key":"claw","count":2,"choice_group":null}] where "bite" and "claw" are other rule_elements keys whose mechanics are attacks - never the multiattack's own key or another multiattack.
 - legendary action: NOT a multiattack. Use mechanic.kind "attack", section "legendary_action", activation.kind "legendary", usage.kind "resource", and costs against the declared legendary pool.
-- standard derivation: strength 16 gives modifier +3; with proficiency_bonus +2, {"skill":"athletics","ability":"strength","value":5,"derivation":"standard","note":null}. If the value you want does not equal ability modifier + proficiency bonus (or + twice proficiency for expertise), set derivation "explicit_override"; never force mismatched arithmetic into "standard".
 """
 
 
@@ -47,11 +46,11 @@ USAGE FIELDS - usage.kind decides which sibling fields may appear:
   manual                         recharge_range null; uses and resource_key optional
 Never set recharge_range on any kind other than recharge.
 
-DERIVED MATH - these are checked arithmetically:
-- challenge.proficiency_bonus must match challenge.rating on the standard 5e table.
-- senses.passive_perception must equal 10 + the Perception skill value when a Perception skill entry exists.
-- Skills and saves with derivation "standard" equal ability modifier + proficiency bonus; "expertise" adds proficiency twice; use "explicit_override" for anything else.
-- vitality.hit_points: method "formula" sets formula and leaves fixed_value null; method "fixed" sets fixed_value and leaves formula null. displayed_average, when set, equals the formula average.
+DERIVED MATH - the server computes; you declare intent:
+- Skills and saves with derivation "standard" or "expertise" are computed from the ability scores and challenge rating you set; the value you emit is advisory and will be overwritten. "explicit_override" is the only derivation whose value is trusted - use it for intentional exceptions.
+- challenge.proficiency_bonus (from rating) and vitality.hit_points.displayed_average (from the dice formula) are likewise server-computed.
+- senses.passive_perception is NOT server-computed: it must equal 10 + the Perception skill value when a Perception skill entry exists.
+- vitality.hit_points: method "formula" sets formula and leaves fixed_value null; method "fixed" sets fixed_value and leaves formula null.
 
 REFERENCES - declare before you reference:
 - Multiattack sequences reference other rule_elements keys; a multiattack may not reference itself or another multiattack.
@@ -83,7 +82,7 @@ ESCAPE HATCH
 1. Exactly one armor class has default=true.
 2. Every usage object matches its kind's row above, and every recharge usage sets recharge_range.
 3. Every key you reference is declared somewhere in the definition.
-4. proficiency_bonus matches CR, and passive_perception matches Perception.
+4. passive_perception equals 10 + the Perception skill value when a Perception entry exists.
 5. Every human_adjudicated mechanic has automation_support "manual".
 """)
     return "\n".join(parts)
--- a/statblocks_v1/domain/derived.py
+++ b/statblocks_v1/domain/derived.py
@@ -12,6 +12,8 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
+from pydantic import ValidationError
+
 from statblocks_v1.domain.primitives import AbilityName, DiceExpression
 from statblocks_v1.domain.profiles import AbilityScores, ProficiencyDerivation
 from statblocks_v1.domain.rule_elements import StatblockDefinitionV1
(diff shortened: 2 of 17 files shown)


Drakosfire (commented):
**Review verdict: stacked shape is correct, but two blockers remain in the incremental layer.**

Centralizing derivation math and importing it from validation is the right architecture: compute and check now share one implementation, explicit overrides retain authored authority, and the generation pipeline applies corrections before the candidate receipt is built.

Blockers:

1. The derivation pass can manufacture a schema-invalid definition after the provider payload has already passed Pydantic validation. `model_copy(update=...)` does not validate updates. A valid provider payload can use `formula={count:1, die:2, modifier:-2}` with `displayed_average=1`; the server computes `-1`, writes it through `model_copy`, then domain validation sees an exact arithmetic match and emits no HP error. The candidate is returned/stored with `displayed_average=-1` despite the contract's `ge=1`. Revalidate the reconstructed definition (and fail closed with diagnostics), or explicitly reject/clamp impossible computed averages. Add an adversarial generate test proving no invalid nested model escapes.
2. This PR bumps runtime provenance to prompt v5 but does not regenerate the published candidate/revise fixtures. They remain at v3 on the stack head. Even after #26 is corrected to v4, this incremental PR must advance them to v5 and keep a test tied to `PROMPT_VERSION`.

Process note: there is currently no Actions run for this head because the workflow only triggers for pull requests targeting `main`. After #26 merges, retarget #27 and run the checks before merge. The manually reported focused lane is useful, but it is not a check attached to this commit.

statblocks_v1/domain/derived.py
```diff
@@ -0,0 +1,136 @@
...
+            adjustments.append(
+                DerivedValueAdjustmentV1(
+                    "vitality.hit_points.displayed_average",
+                    hit_points.displayed_average,
+                    average,
+                )
+            )
+            hit_points = hit_points.model_copy(update={"displayed_average": average})
+
+    if not adjustments:
```
Drakosfire, line 131: **Blocking — unvalidated `model_copy` can create a contract-invalid candidate.** `DiceExpression.modifier` is unbounded while `displayed_average` is `ge=1`. A provider payload with `1d2-2` and emitted average `1` parses successfully; this computes `-1`, and `model_copy(update=...)` accepts it without validation. The later domain check sees the values agree, so the API can return/persist `displayed_average=-1`. Revalidate the full derived definition or fail closed when a computed value violates the schema, with an adversarial service test.

statblocks_v1/application/prompts.py
```diff
@@ -4,15 +4,14 @@
 from statblocks_v1.application.commands import GenerateStatblockCommandV1, ReviseStatblockCommandV1
 from statblocks_v1.domain.rule_elements import StatblockDefinitionV1
 
-PROMPT_VERSION = "statblock-generation-prompt-v4"
+PROMPT_VERSION = "statblock-generation-prompt-v5"
```
Drakosfire, line 7: **Blocking provenance drift in the stack.** This advances runtime receipts to prompt v5, but neither published API response fixture is regenerated; both still say v3. #26 must first fix them to v4, and this PR must then advance them to v5. Add a fixture test against `PROMPT_VERSION` so a version bump cannot land without its consumer artifacts.

Drakosfire:
Review addressed in `e6d906b` (rebased onto the fixed base; branch force-pushed):

1. **Invalid model after derivation** — real find, thank you. `DiceExpression.modifier` is unbounded, so `1d2-2` averages −1 while `displayed_average` requires ≥ 1, and `model_copy(update=...)` skips validation. Two guards now: averages below the field floor are never written (authored value stands, and `HP_DISPLAYED_AVERAGE_MISMATCH` keeps the pathological formula visible on the receipt), and the derived model is fully revalidated with a fallback to the provider's original definition — compute can never make the candidate less valid than the v4 behavior. New domain + service tests pin both (`1d2-2` → candidate stores 1, receipt flags the mismatch).
2. **Fixture provenance** — fixtures regenerated at prompt v5; the provenance assertions added on the base branch now enforce this mechanically.

Lane: 301 passed. Branch remains based on #26's head; retarget to `main` after #26 merges.

statblocks_v1/domain/derived.py
```diff
@@ -0,0 +1,151 @@
...
+
+    hit_points = definition.vitality.hit_points
+    if hit_points.formula is not None:
+        average = hit_point_average(hit_points.formula)
+        # displayed_average is bounded (>= 1) while DiceExpression.modifier is
+        # not: a pathological formula (1d2-2) can average below the field's
+        # floor. Skip the write; the authored value stays and the validator's
+        # mismatch error keeps the weird formula visible instead of storing a
+        # contract-violating number.
+        if average >= 1 and hit_points.displayed_average != average:
```
Drakosfire, line 121: **Blocking — `displayed_average: null` makes the pathological formula silent.** For `1d2-2`, this guard skips the write. When the provider supplied `null`, `_diagnostics`/domain validation does not emit `HP_DISPLAYED_AVERAGE_MISMATCH` because that validator is conditional on `displayed_average is not None`. The candidate therefore keeps a non-representable HP formula with no issue, despite the comment claiming the mismatch remains visible. Add an explicit validation/failure signal for computed averages below 1 and cover the `null` path end-to-end.

Drakosfire (commented):
## Stack Review 2 — one blocker remains

The original contract-violation finding is mostly fixed: the server no longer writes a derived HP average below the Pydantic floor, the rebuilt definition is fully revalidated, and published fixtures now advance to prompt v5.

One edge still contradicts the new safety claim. For a valid provider payload with `formula = 1d2-2` and `displayed_average = null`, this branch skips the write and returns no adjustment. Domain validation only checks `HP_DISPLAYED_AVERAGE_MISMATCH` when `displayed_average is not None`, so the candidate receives no HP diagnostic at all. The comment says the validator keeps the pathological formula visible, but that is only true for an authored non-null average such as `1`.

Please make an impossible computed average explicit regardless of whether the provider supplied an average. Viable fixes include: reject/constrain formulas whose computed average is below 1, emit a domain validation issue whenever the formula average is below the representable field floor, or fail generation closed for this state. Add the end-to-end service test with `displayed_average: null` and assert the chosen error signal is present. The current unit test includes `None` but only asserts that no write occurs, which misses the silent-clean-candidate behavior.

Drakosfire:
Remaining blocker addressed in `9899d11`:

The null path was silent because the defect had been framed as "displayed average vs. formula," but the actual defect is the formula itself. The validator now checks it directly: whenever a typed formula averages below 1, **`HP_FORMULA_AVERAGE_INVALID`** fires (error, path `vitality.hit_points.formula`) — independent of `displayed_average` null-ness. For the pathological case it subsumes `HP_DISPLAYED_AVERAGE_MISMATCH` (one root-cause error, not two); sane formulas keep the existing mismatch semantics unchanged.

Your exact scenario end-to-end: `1d2-2` + `displayed_average: null` → derivation skips the write, candidate stores `null`, receipt carries `HP_FORMULA_AVERAGE_INVALID`, Workbench operator sees the formula flagged. New coverage: validator unit tests for the null case, the subsumption case, and the unchanged sane-formula mismatch; service-level test with `displayed_average = null` asserting the receipt signal; the existing `displayed_average = 1` service test updated to the new code.

`VALIDATOR_VERSION` bumped to v1.2 and all six receipt-carrying fixtures regenerated (revise composite now `v1.2+statblock-key-preservation-v1`). Lane: 305 passed.

Drakosfire (commented):
## Stack Review 3 — no blocking code findings

The Review 2 null-path defect is resolved correctly in `9899d11`:

- Pathological formulas are now validated independently of whether `displayed_average` is present.
- `HP_FORMULA_AVERAGE_INVALID` is emitted at `vitality.hit_points.formula` and blocks persistence.
- The impossible derived value is still not written into the candidate.
- The service lane now covers both authored `displayed_average=1` and `displayed_average=null`.
- Sane formulas retain the existing `HP_DISPLAYED_AVERAGE_MISMATCH` behavior.
- `VALIDATOR_VERSION` was bumped and all six receipt-bearing API fixtures were regenerated.

I found no new blocking code issue in this commit.

Non-blocking wording correction: `"it cannot yield a positive HP value"` is stronger than the invariant actually checked. A formula such as `1d6-3` has a floored average of 0 but can roll a positive result. The accurate explanation is that the formula's computed average is below the contract minimum of 1.

**Merge-readiness requirement:** #26 is now merged, but this PR still targets `feat/statblocks-v1-prompt-and-schema-guidance`. Retarget it to `main` and rebase/merge current `main` into the branch before merging. The current head is one merge commit behind `main` and has no attached Actions run because its base is not `main`. After retargeting, verify the diff remains the intended incremental 17-file layer and run the checks.

Review count: **Stack Review 3**.

PR #31, opened 2026-09-04
E2 cutover: consume GenerationClient and retire MODEL_POLICY

Drakosfire:
## Summary

- Paired GenerationEngine PR #3 merged as `0d01547e2d9afec68e87b4c8f7e6aaa047e8c42a`; this PR pins that exact accepted merge in both `pyproject.toml` and `uv.lock`.
- Migrates every inventoried GenerationEngine consumer (MapSpec, SVG mask, PCG, map T2I, map inpaint, card item/core/i2i, user image selector) onto `GenerationClient`.
- Product-owned action → generic profile mapping lives in `shared/inference_policy.py`. Explicit catalog IDs preserve measured behavior where a profile default would silently change the model.
- GenerationEngine returns image bytes. DungeonMindServer validates and publishes them through `upload_generated_image_bytes`, registers the asset, and returns the product URL.
- Statblocks v1 keeps its `DefinitionProvider`, receipt, and worker-thread seams. Its GE adapter maps failures and carries observed provider/model/request/response/token/latency truth into the existing receipt. Default model resolution belongs to the GE profile; an explicit model remains only as a provider-neutral override.
- Product lease math uses one 90-second inference budget plus asset budget and margin. It contains no knowledge of GE retry counts.
- `MODEL_POLICY.json` and the old direct Statblocks OpenAI provider are deleted.

## Behavioral parity

| Action | Accepted behavior |
| --- | --- |
| MapSpec | `structured_low_cost` → `gpt-5.1` |
| SVG mask | `text_fast` → `gpt-5.1` |
| PCG | `text_fast` → `gpt-5.1`; unknown token values remain `None` |
| Card item | `structured_low_cost` + explicit `gpt-4o` |
| Statblock v1 | `structured_high_reliability`; GE resolves `gpt-5.6-luna` by default; 90s overall inference budget |
| Map T2I | explicit `gpt-image-1.5` |
| Map inpaint | `edit_image` + explicit `gpt-image-1.5` |
| Card core image | explicit `nano-banana-pro` |
| Card I2I | explicit `flux-lora-i2i`, strength 0.85 |
| User image selector | product aliases map to explicit catalog IDs |

## Final-pin evidence

- [x] `uv lock --check`
- [x] Focused cutover suite: 22 passed
- [x] Broad Python 3.12 suite: 330 passed, 5 deselected, 27 warnings
- [x] Exact dependency source and lock both resolve to GE merge `0d01547e2d9afec68e87b4c8f7e6aaa047e8c42a`
[...]

PR #32, opened 2026-09-11
CUTOVER: retire duplicate card provider bypass

Drakosfire:
## Gate G finding

Settling Gate G found that the unmounted monolithic CardGenerator router is still imported at runtime for `CardSessionData`. It therefore still instantiated a direct OpenAI client and retained a second `/generate-item-dict` implementation for the same card structured-generation action migrated to `CardGenerationService` in DMS #31.

## Change

- delete the dormant duplicate endpoint, schema, and import-time OpenAI client;
- keep the mounted compatibility route pointing at the authoritative `CardGenerationService` path;
- add an architecture fitness regression that rejects direct OpenAI/Fal imports in both the authoritative card service and the still-imported monolithic surface.

This is an owning-runtime fix discovered by Gate G. It does not change OverMind authority and does not broaden into other explicitly named direct-provider debt.

## Evidence

- `UV_PYTHON=3.12 uv run --python 3.12 pytest -q tests/test_generationengine_cutover_fitness.py tests/test_redteam_hardening.py tests/test_packaging_identity.py`
  - 29 passed
- `git diff --check`
  - pass

After this merges, Gate G will advance its accepted DMS revision to the actual merge SHA and rerun affected evidence/audits before C4 changes state.


Drakosfire (commented):
## Review Cycle 1 — APPROVE / MERGE-READY

Formal judgment against exact head `996fe62`.

- PASS: removes the direct OpenAI client and duplicate card structured-generation route discovered by Gate G.
- PASS: the mounted current and compatibility routes continue to converge on `CardGenerationService` → `GenerationClient`.
- PASS: regression coverage now checks both the authoritative service and the still-imported monolithic module for direct OpenAI/Fal bypasses.
- PASS: focused local evidence: 29 passed.
- PASS: GitHub red-team hardening is green on this exact head.
- PASS: no OverMind authority or unrelated direct-provider debt is changed.

No blocker found. Merge this owning-runtime fix, then rerun Gate G against its actual merge SHA.

